PR #2, opened 2019-07-29
Allow templates to show up in multiple categories

andymckay:
We'll get a language returned by the language detection on the repo. However some templates work in more than one case or language, for example the `C/C++ with Make` template. This changes category to an array so that we can have a template show up in more than one place and makes some guesses when this would be appropriate.

Supports github/dreamlifter#953


mscoutermarsh (approved):
💯

PR #17, opened 2019-08-08
Go template - use->uses

cdb:
Fixes a small bug in the template, `use` should be `uses`


ahmetb:
This change has still not rolled out apparently.

mscoutermarsh:
@ahmetb Changes being synced with github.com in next couple hours.

PR #19, opened 2019-08-09
Test node in all current versions

wesleytodd:
Node packages should be tested in all current versions by default.  We are working on setting some guidelines as mentioned in https://github.com/actions/setup-node/issues/25, so until aliases are supported we would want to keep this list up to date with the node LTS plan: https://nodejs.org/en/about/releases/

As I am not in the beta, I do not have the ability to try these out, but I wanted to make sure we setup people up with the best practice from day one, so figured I would get this in early.  I copied from the python example, so think it should work but am happy to make any changes necessary to get this merged.

### DIFF
--- a/ci/node.js.yml
+++ b/ci/node.js.yml
@@ -7,12 +7,16 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        node-version: [8.x, 10.x, 12.x]
+
     steps:
     - uses: actions/checkout@master
-    - name: Use Node.js 10.x
+    - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v1
       with:
-        version: 10.x
+        version: ${{ matrix.node-version }}
     - name: npm install, build, and test
       run: |
         npm install


damccorm:
This generally LGTM. My only big question is should we do this as the only template or as a separate template (and keep the non-matrixed version)? If we do both, which should be the default node template?

I'm pretty non-opinionated on this one, @chrispat might have more thoughts?

@wesleytodd thanks for the quick follow through on this!

wesleytodd:
Hopefully the point of making this the only one is that it is the highly recommended default for packages.  If you are also thinking about service testing then I could see having a secondary template, but it seems like OSS packages are the most important for this.

chrispat:
@wesleytodd and @damccorm  would it make more sense here to have LTS and LATEST as the options in this matrix given the goals?

damccorm:
Eventually, we still need to add those options though. https://github.com/actions/setup-node/issues/26

damccorm:
That should be pretty quick to implement when the time comes, just want to make sure we're getting the options right though before we do the work and people start to rely on them.

wesleytodd:
Is there anyway one of you could get me into the beta so I can start using these?  We talked in [the last ExpressJS TC meeting](https://github.com/expressjs/discussions/issues/85) about getting more automation running, and now that you announced this I think we would prefer to keep that all on github.

jeremyepling:
@wesleytodd have you signed up expressjs for the beta? https://github.com/features/actions/signup/?account=expressjs

After that I can get you in.

wesleytodd:
I was not aware you could register on behalf of an organization, so I had just done it as myself :)

Thanks!

chrispat:
you should be enabled now

PR #24, opened 2019-08-12
Fix arguments for go-sample workflow in step checkout

mmoscher:
Set argument `path` for action using `with` as documented.

https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepswith

### DIFF
--- a/ci/go.yml
+++ b/ci/go.yml
@@ -15,7 +15,8 @@ jobs:
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@master
-      path: ${{ go.module-path }}
+      with:
+        path: ${{ go.module-path }}
 
     - name: Get dependencies
       working-directory: ${{ go.module-path }}


ipfans:
I still have a question that as https://github.com/ipfans/action-demo/commit/26136ab8102331601e32ef87ae17940bc38b51b8/checks, it seems that virtual env ${{ go.module-path }} should setup by me? or which value it should be?

mmoscher:
I'm not quite sure what ` ${{ go.module-path }}` refers to. Didn't found any output/option on actions/setup-go which is named `module-path`. However, deleting these expressions entirely in the workflow, i.e. as you did as well, worked for me. But this shouldn't (IMHO) be part of this PR.

rodrigo-brito:
I'm having the same problem with `${{ go.module-path }}`. Actions reporting `Unrecognized named-value: 'go'`

mmoscher:
Exactly, same for me.

rodrigo-brito:
@mmoscher the `setup-go` repo don't use it in the example. I removed and work well to me: https://github.com/rodrigo-brito/gocity/blob/master/.github/workflows/go.yml

mscoutermarsh:
Thanks all! Currently `setup-go` doesn't have module-path as an output. Removing it from the template in another PR.

PR #30, opened 2019-08-13
Create workflow for publishing gems

pmarsceill:
Adds a workflow to publish Ruby Gems to RubyGems.org and GPR. cc @mscoutermarsh


mscoutermarsh:
Can you add a properties file? First category should be "Ruby"

PR #33, opened 2019-08-14
Update main.yml

nageswar255:


damccorm:
I don't think we need this, this is just a starter template to get people going and doesn't add any value to this repo IMO. Closing, please comment if you disagree and we can reexamine

PR #34, opened 2019-08-14
Update nodejs.yml

nageswar255:


damccorm:
I don't think we want this - see conversation in https://github.com/actions/starter-workflows/pull/19 on why we have the node template as it is. I'm going to close, feel free to comment if you disagree and we can reopen as appropriate

PR #37, opened 2019-08-14
Update c-cpp.yml for configure/automake

rurban:
make test is a perl'ism (and probably other), autotools use make check.
add make distcheck

### DIFF
--- a/ci/c-cpp.yml
+++ b/ci/c-cpp.yml
@@ -13,5 +13,7 @@ jobs:
       run: ./configure
     - name: make
       run: make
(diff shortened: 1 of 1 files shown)


lilianmoraru:
Most C/C++ projects use CMake.
Should there also be a CMake configuration?
How would one go about creating one, so it does not conflict with this configuration?

rurban:
@lilianmoraru: true, with cmake make test would be better. I have no idea how to detect cmake in this limited yaml framework.

PR #47, opened 2019-08-16
Add a starter cmake workflow

jherico:
Many modern C/C++ projects are moving away from pure configure/make style builds and towards CMake builds for easier cross platform manageability.   There should be a simple starter CI workflow for such projects.

Icon is from https://commons.wikimedia.org/wiki/File:Cmake.svg and is licensed under Creative Commons Attribution 2.0 Generic license.


ci/cmake.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v1
+    - name: configure
+      run: cmake .
```
podusowski: You should rather avoid in-source builds. Some projects won't even allow it.

jherico: While I agree with the sentiment, determining and creating a build directory in a cross platform way seems problematic.  I don't want to have the starter workflow have a bunch of per-platform logic to calculate a suitable build directory name, since I'd then have to worry about the chosen directory name conflicting with an existing directory from the checkout.  

If you have any suggestions on how to resolve this in a platform-neutral way that doesn't end up obscuring the cmake specific content, I'd be happy to incorporate them.

DeveloperPaul123: Isn't it pretty common to build in a "build" subdirectory of the source? I've seen that quite a bit. I've used that myself, but I've also used the case where the build directory is at the same level as the source file. Consider the project `foo`:

````
foo
|
|-foo-src
  |-CMakeLists.txt
|-foo-build
|-foo-build-special-version
````

I think these two cases cover the majority of cases in how CMake is typically configured/used. Have you observed any other patterns?

Comment (jherico) [reply ci/cmake.yml]: I've discovered the `cmake -E make_directory` command provides a cross platform way to make directories.  I will update this PR to reflect the common build practice.

Comment (DeveloperPaul123) [reply ci/cmake.yml]: Can't you just specify the build directory at configuration time? 

````cmake
cmake -S <source> -B <build> ...<other flags>
````
This should create the build directory if it doesn't exist. Or am I wrong?

**Comment by jherico** [comment]
Apparently so.

File: ci/cmake.yml (hunk 1 above)
Comment (jherico) [reply ci/cmake.yml]: The `-S <source> -B <build>` syntax doesn't work with CMake 3.12, which is what's on at least some of the build hosts.  However, the make directory mechanism will work.

Comment (DeveloperPaul123) [reply ci/cmake.yml]: Ah I see, didn't realize we're working with CMake 3.12.

Comment (jherico) [reply ci/cmake.yml]: Me either, which is why I nearly lost my mind trying to debug why the -S / -B stuff wasn't working.

Comment (DeveloperPaul123) [reply ci/cmake.yml]: Thanks for handling it though, looking forward to using this action for my cmake projects!

**Comment by rigtorp** [comment]
Hmm, I made it quite a bit simpler and also with windows support:
```
name: C/C++ CI

on: [push]

jobs:
  build-ubuntu:

    runs-on: ubuntu-latest
    
    steps:
    - uses: actions/checkout@v1
    - name: configure
      run: mkdir build && cd build && cmake -DCMAKE_CXX_FLAGS="-Werror" ..
    - name: build
      run: cmake --build build
    - name: test
      run: cd build && ctest

  build-windows:

    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [windows-latest, windows-2016]
    
    steps:
    - uses: actions/checkout@v1
    - name: configure
      run: mkdir build && cd build && cmake ..
    - name: build
      run: cmake --build build --config Debug
    - name: test
      run: cd build && ctest
```

File: ci/cmake.yml [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v1
+    - name: create-build-dir
```
ethomson: Let's use more descriptive names here, eg `name: Create Build Environment`

ci/cmake.yml
```diff
@@ -0,0 +1,22 @@
...
+      run: cmake -E make_directory ${{runner.workspace}}/build
+    - name: cmake-configure
```
ethomson: Let's use more descriptive names here, eg `name: Configure CMake`

ci/cmake.yml
```diff
@@ -0,0 +1,22 @@
...
+      working-directory: ${{runner.workspace}}/build
+      run: cmake $GITHUB_WORKSPACE
+    - name: build
```
ethomson: Let's capitalize this, eg `Build`.

ci/cmake.yml
```diff
@@ -0,0 +1,22 @@
...
+      working-directory: ${{runner.workspace}}/build
+      run: cmake --build . --config Release
+    - name: test
```
ethomson: Let's capitalize this, eg `Test`.

ethomson:
One question: why not just use the default config instead of specifying `Release`?  I usually do debug builds for my CI, and I think that's common to only do release builds for deploy?

jherico:
@ethomson I wanted to illustrate how to ensure a specific build type was done.  However, I've updated the build type to be environment variable driven now with a comment to that effect in the worflow.  Hopefully this is clearer, and allows a client to more easily switch to the kind of build they want depending on circumstance.

ethomson:
Thanks @jherico!  We have a little work we're doing to improve the starter workflows section, and I want to wait for that to land before I merge this.  It will help the organization overall.  But we'll land this 🔜 .

Aidoboy:
@ethomson Is there any ETA?

rpavlik:
also, any chance the Windows runners can get something newer than 3.6? (or the docs be updated if that's not accurate, though the way FindOpenGL.cmake is acting, I think it is accurate...)

jherico:
@rpavlik the cmake version on `windows-latest` is actually currently 3.16.0 per my latest test.  see the "dump cmake version" step results here:  https://github.com/jherico/actions-test/runs/346077993

rpavlik:
Oh interesting @jherico , thanks for checking that. The docs are mega out of date then, and I must have been doing something else weird with opengl.

jherico:
@rpavlik I think the page is somewhat confusing, because on some platforms CMake is listed multiple times.  The Windows platform correctly lists CMake 3.16, but the Ubuntu platform lists both CMake 3.12 and 3.6.  I suspect that this has to do with the bundled cmake that comes in the Android SDK/NDK being reported along with the main system CMake.

cristianadam:
I've blogged about building C++ project using CMake and GitHub Actions: https://cristianadam.eu/20191222/using-github-actions-with-c-plus-plus-and-cmake/

I have a HelloWorld project https://github.com/cristianadam/HelloWorld/ which I used as an example.

I used CMake as a scripting language to download the latest CMake, and Ninja. Then configure, build, run tests, upload artifacts, and then handle releases.

WSudul:
Hello,
I've looked at the PR and related blog post and this looks like useful thing.
Additionally someone has already used this idea in their repositories ( acclimate/acclimate repo in comments above).
What is blocking the PR from being merged? All (?) comments were addressed, right?

zbeekman:
I think that this could be further simplified with modern CMake since it can create the build directory for you and be more shell/OS agnostic. Also, other compiled languages, like Fortran can use CMake build systems.

I may work up an alternate updated PR with my simplifications unless anyone objects.

kukulkankanpull (commented):
ADD CHANGES

ci/cmake.yml
Safran1022: https://commons.wikimedia.org/wiki/File:Cmake.svg

ci/cmake.yml
Safran1022: https://commons.wikimedia.org/wiki/File:Cmake.svg

ci/cmake.yml
Safran1022: https://commons.wikimedia.org/wiki/File:Cmake.svg

ci/cmake.yml
Safran1022: https://commons.wikimedia.org/wiki/File:Cmake.svg

PR #54, opened 2019-08-20
Add Scala workflow

JackTreble:
Added simple Scala starter template along with properties file.
Need to find a svg scala logo

### DIFF
--- a/ci/properties/scala.properties.json
+++ b/ci/properties/scala.properties.json
@@ -1,0 +1,6 @@
+{
+    "name": "Scala",
+    "description": "Build and test a Scala project with SBT.",
+    "iconName": "scala",
+    "categories": ["Scala", "Java"]
+}
--- a/ci/properties/scala.properties.json
+++ b/ci/properties/scala.properties.json
@@ -1,6 +1,6 @@
 {
     "name": "Scala",
     "description": "Build and test a Scala project with SBT.",
-    "iconName": "scala",
+    "iconName": "blank",
     "categories": ["Scala", "Java"]
 }
--- a/ci/scala.yml
+++ b/ci/scala.yml
@@ -1,0 +1,15 @@
+name: Scala CI
+
+on: [push]
+
+jobs:
+  build:
+ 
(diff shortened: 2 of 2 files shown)


hrj:
How does this work? How does Java and SBT get installed?

sim642:
I made my own Scala SBT workflow based on the existing Java Maven one: https://github.com/sim642/adventofcode/commit/86b9fd9af1140a1d1a712b61e5a78c3475a98c46.

`actions/setup-java` should work for more explicitly getting Java and controlling the version. I'm guessing SBT comes from the [standard installed software](https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions) (as part of Scala?). It should be respecting `sbt.version` in `project/build.properties`.

JackTreble:
@hrj sbt gets downloaded when the first sbt command is executed
```
Run sbt compile
  sbt compile
  shell: /bin/bash -e {0}
Downloading sbt launcher for 1.2.8:
  From  https://repo.scala-sbt.org/scalasbt/maven-releases/org/scala-sbt/sbt-launch/1.2.8/sbt-launch-1.2.8.jar
    To  /home/runner/.sbt/launchers/1.2.8/sbt-launch.jar
/home/runner/.sbt/launchers/1.2.8/sbt-launch.jar: OK
 Getting org.scala-sbt sbt 1.2.8  (this may take some time)...
```
like sim642 said i expect its part of Scala from the standard installed software.

@sim642 good Idea on adding in the `Set up JDK 1.8` action, I originally based this off the [clojure](https://github.com/actions/starter-workflows/blob/master/ci/clojure.yml) ci which does not do this.
I will make this change

sullis:
I created a slightly different solution.  Here is my PR:

https://github.com/actions/starter-workflows/pull/128

(sbt.yml)

JackTreble:
I agree, as far as i can tell there is no specific Java icon so i have set it to blank

andymckay:
Thank you @JackTreble

PR #55, opened 2019-08-20
update Erlang workflow to rebar3

tsloughter:
[rebar](https://github.com/rebar/rebar) has been deprecated for a few years now and [rebar3](https://github.com/erlang/rebar3) has taken its place.

Since rebar3 no longer has a separate step for fetching dependencies so this step is removed. Common Test is enabled in the test run so it will run all tests.

The Erlang docker image is used instead of ubuntu because ubuntu has no package for rebar3.

### DIFF
--- a/ci/erlang.yml
+++ b/ci/erlang.yml
@@ -8,11 +8,12 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    container:
+      image: erlang:22.0.7
+
     steps:
     - uses: actions/checkout@v1
-    - name: Install dependencies
-      run: rebar get-deps
     - name: Compile
-      run: rebar compile
+      run: rebar3 compile
     - name: Run tests
-      run: rebar skip_deps=true eunit
+      run: rebar3 do eunit, ct


jclem:
@tsloughter Thanks for noticing this. I'm not sure how to approach it yet—using non-Docker actions is typically easier (since the container form forces you to either run every step in the container or specify the container for every Erlang step)  and a bit faster.

Instead, I will probably wait until I have a good solution for [setup-erlang](https://github.com/actions/setup-elixir/issues/2), and then include that action in this starter workflow.

tsloughter:
Ah ok, hm, so the Elixir one, which uses a docker image like this, will change to a non-Docker action as well?

jclem:
@tsloughter Went ahead and merged for now—thanks for the PR. I'll come back and re-implement this w/setup-erlang when I've got the time.

tsloughter:
@jclem cool, thanks, looking forward to trying the setup-erlang verison. I hope to use this for relx and possibly rebar3 if it goes well with relx, once the organization is accepted into thhe beta.

PR #56, opened 2019-08-20
fix npm-publish.yml

zachguo:


zachguo:
also it doesn't seem to make a lot of sense to publish package on every pull request.

ci/npm-publish.yml
```diff
@@ -42,6 +39,7 @@ jobs:
           node-version: 12
           registry-url: https://npm.pkg.github.com/
           scope: '@your-github-username'
+      - run: npm ci
```
denlb: Why is this necessary?

zachguo: `node_modules` was absent if I didn't run `npm install`.

denlb: It is absent but does it fail when you publish?

zachguo: I did run into errors when I tried to set Actions up for two of my projects.
Has `setup-node` been changed? It could be fixed by changing somewhere else.
I ran into a 500 error when I tried to publish to GPR but now it seems to be fixed.

zachguo: BTW `npm publish` would run `npm prepare` which may require `node_modules` or whatever processes a user defines.

zachguo: I just checked my Action logs, it failed at `npm prepare` where `webpack` tried to create a package build.

denlb: make sense then but I don't see it necessary for an easy setup - not every package has a prepare script

zachguo: I get your point, adding `npm ci` by default may generate unnecessary data transfer. Do you want me to remove this line?
And I think it's worthwhile to be added to docs, it's probably a common hurdle people may run into.

andymckay:
This looks good, thanks @zachguo

PR #69, opened 2019-08-25
Publish Python package to PyPI workflow

montudor:
Publishing a Python package directly to [PyPI](https://pypi.org/) when you create a release is a really common use case for CI. I think including this as a starter workflow will be very useful for when GitHub Actions is released to everyone.

### DIFF
--- a/ci/python-publish.yml
+++ b/ci/python-publish.yml
@@ -1,6 +1,8 @@
 name: Upload Python Package
 
-on: [release]
+on:
+  release:
+    types: [created]
 
 jobs:
   deploy:
@@ -8,17 +10,14 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Set up Python
-      if: github.event.action == 'created'
       uses: actions/setup-python@v1
       with:
         python-version: '3.x'
     - name: Install dependencies
-      if: github.event.action == 'created'
       run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel twine
     - name: Build and publish
-      if: github.event.action == 'created'
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
--- a/ci/python-publish.yml
+++ b/ci/python-publish.yml
@@ -1,0 +1,27 @@
+name: Upload Python Package
+
+on: [release]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python
+      if: github.event.action == 'created'
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      if: github.event.action == 'created'
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
+      if: github.event.action == 'created'
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*
--- a/ci/properties/python-publish.properties.json
+++ b/ci/properties/python-publish.properties.json
@@ -1,0 +1,6 @@
+{
+    "name": "Publish Python Package",
+    "description": "Publish a Python Package to PyPI on release.",
+    "iconName": "python",
+    "categories": ["Python"]
+}


ci/python-publish.yml
```diff
@@ -0,0 +1,27 @@
...
+        python-version: '3.x'
+    - name: Install dependencies
+      if: github.event.action == 'created'
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
+      if: github.event.action == 'created'
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
```
Julian: Ideally this should use the new [token functionality](https://pypi.org/help/#apitoken)

ci/python-publish.yml
```diff
@@ -0,0 +1,27 @@
...
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
```
Julian: And I think here ideally this could use PEP517 builders so that it works regardless of whether someone's using setuptools or not. I.e. `pep517.build .`

ci/python-publish.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python
+      if: github.event.action == 'created'
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
```
Julian: This presumably needs to be configurable, right?

ci/python-publish.yml
```diff
@@ -0,0 +1,27 @@
+name: Upload Python Package
+
+on: [release]
```
chrispat:
```suggestion
on:
  release:
    types: [created]
```
This sill make sure this workflow only runs when the release is created and not on any of the other events. Then remove all of the conditionals on the steps.

montudor: Thanks! I was looking for a better way to do this but didn't notice this in the actions documentation.

ci/python-publish.yml
montudor: Though I agree using tokens is a better method, having the option to configure the username and password via secrets allows the user to choose between using their credentials or a token, depending on their use case.

ci/python-publish.yml
montudor: Yeah, it is up to the user to change this when they generate the workflow from this template. I chose to set the default as Python 3 since Python 2 support is getting dropped soon.

ci/python-publish.yml
montudor: I think it's best to stick to the current command that the official documentation on [packaging python projects](https://packaging.python.org/tutorials/packaging-projects/) uses.

Julian: Yep makes sense.

pradyunsg:
Would y'all be willing to recommend using https://github.com/pypa/gh-action-pypi-publish here?

That was built precisely because we'd want to have "one good way" to do uploads (and want to subtly encouraging users to use tokens, a more secure default). For what it's worth, there's no v1.0.0 tag on that action yet, but master is good to go for now.

@webknjaz could you create a `v1.0.0` tag on that repo, so that it can be used as `uses: pypa/gh-action-pypi-publish@v1`?

montudor:
@pradyunsg according to the [contribution guidelines](https://github.com/actions/starter-workflows/blob/master/CONTRIBUTING.md):
> Cannot use an Action that isn't in the actions organization

So I doubt the action will be allowed.

The current workflow allows the use of tokens and username/passwords. I set it up this way to provide as much flexibility as possible, though, I understand tokens are best practice so maybe this would be a good change to the current workflow.

@webknjaz as for using PEP517, you'll have to create a pull request and see what the devs think about this change. I personally used the method listed in the official [Python documentation](https://packaging.python.org/tutorials/packaging-projects/) but I can understand your concerns.

pradyunsg:
*sigh* Thanks for the prompt response @montudor!

Well, sounds like a lot of folks are gonna roll out their own PyPI-publish bit for GitHub Actions. :(

None the less, I do think at least a comment nudging users to "use API tokens if you can" would be a good addition, should such an addition be accepted.

webknjaz:
@montudor to fix the token concern, it's enough to change `TWINE_USERNAME`'s value to `__token__` and point to the PyPI.org docs about those I think, explaining that it's an officially recommended way to go. Ref: https://github.com/pypa/warehouse/issues/6211

montudor:
@webknjaz I completely agree, I'm sure the devs will accept a PR with this change.

webknjaz:
@montudor do you think you could at least point to PyPA guides? It's a collection of guides that are collectively maintained by the developers or mainstream packaging bits in Python ecosystem (twine/setuptools/pip). And it's under the official Python domain: https://packaging.python.org.

montudor:
@webknjaz I think that's also a good idea.

pganssle:
> I personally used the method listed in the official Python documentation but I can understand your concerns.

We should probably go ahead and remove that from the documentation. It seems that literally saying nothing might be better than what we have now.

Issue #59, opened 2019-08-21
Dockerfile 


pascalandy:
I just tested the Dockerfile CI example and it's working good :) 
Now to go all in, there are many step we need to put in place.

Here is my [.travis.yml ](https://github.com/firepress-org/ghostfire/blob/master/.travis.yml).

### define vars

```
before_script:
# define our labels and tags
  - GHOST_VERSION_DOCKERFILE="$(cat Dockerfile | grep GHOST_VERSION= | head -n 1 | grep -o '".*"' | sed 's/"//g')"
  - GHOST_LATEST_RELEASE="$(git ls-remote --tags https://github.com/TryGhost/Ghost.git | cut -d$'\t' -f2 | grep -E '^refs/tags/[0-9]+\.[0-9]+' | cut -d/ -f3 | sort -rV | head -n1)"
  - GHOST_CLI_LATEST_RELEASE="$(git ls-remote --tags https://github.com/TryGhost/Ghost-CLI.git | cut -d$'\t' -f2 | grep -E '^refs/tags/[0-9]+\.[0-9]+' | cut -d/ -f3 | sort -rV | head -n1)"
  - GHOST_VERSION=${GHOST_VERSION_DOCKERFILE}

  - DOCKER_USERNAME="devmtl"
  - DOCKER_REPO_NAME="ghostfire"
  - CONTAINER_NAME="ghostuat"

  - THIS_DATE=$(date +%Y-%m-%d)
  - THIS_SHA_SHORT=$(git rev-parse --short HEAD)

  - IMAGE_SHA_SHORT=${DOCKER_USERNAME}/${DOCKER_REPO_NAME}:${GHOST_VERSION}-${THIS_SHA_SHORT}
  - IMAGE_VERSION=${DOCKER_USERNAME}/${DOCKER_REPO_NAME}:${GHOST_VERSION}
  - IMAGE_STABLE=${DOCKER_USERNAME}/${DOCKER_REPO_NAME}:stable
```

### install:

```
  - echo "install microscanner Aquasec Scanner ==========>"
  - wget --retry-connrefused --waitretry=1 --read-timeout=5 --timeout=5 --tries=10 https://raw.githubusercontent.com/lukebond/microscanner-wrapper/master/scan.sh && chmod +x scan.sh

  - git clone https://github.com/docker-library/official-images.git ~/official-images
  - echo "Bypass timezone'stest "utc" as we use America/New_York timezone in our image ==========>"
[...]

PR #71, opened 2019-08-26
Template: docker_build_ci.yml

pascalandy:
Following https://github.com/actions/starter-workflows/issues/59.

I had a hard time to understand the syntax of a real life CI scenario looking at the [actual docker template](https://github.com/actions/starter-workflows/blob/master/ci/docker-image.yml).

After many hours reading docs, looking at other repo using actions I found a decent way to bring the CI to life (and on part with Travis CI).

With this template, people will be able to quickly modify the yaml and make it their own.

### Variable

We can see that I fetch VARs from the Dockerfile itself. If you guys prefer, I could hardcode them all if it's easier for everyone to consume the template.

Cheers!

### DIFF
--- a/ci/docker_build_ci.yml
+++ b/ci/docker_build_ci.yml
@@ -1,0 +1,77 @@
+name: docker_build_ci
+on:
+  push:
+  # everyday at 3 am
+  schedule:
+    - cron:  '0 3 * * *'
+
+jobs:
+
+  Job1:
+    name: Job 1/2
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@master
+
+      - name: A) Define VARs
+        run: |
+          echo "We need five information to build our tags:"
+          echo " "
+          echo "devmtl" > DOCKERHUB_USER
+          cat Dockerfile | grep APP_NAME= | head -n 1 | grep -o '".*"' | sed 's/"//g' > APP_NAME
+          cat Dockerfile | grep VERSION= | head -n 1 | grep -o '".*"' | sed 's/"//g' > VERSION
+          date -d "-4 hours" "+%Y-%m-%d_%HH%Ms%S" > DATE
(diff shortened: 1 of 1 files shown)


1138-4EB:
Ref #96

ci/docker_build_ci.yml
```diff
@@ -0,0 +1,77 @@
+name: docker_build_ci
+on:
+  push:
+  # everyday at 3 am
+  schedule:
+    - cron:  '0 3 * * *'
```
andymckay: We don't put `schedule` in our starter workflow, preferring either `push` or `release`.

pascalandy:
@andymckay sure, I understand. It's all about giving a decent template to build Dockerfiles.

What are your specs for a good template? IMHO, at the moment, the template is not useful at all.

1138-4EB:
> What are your specs for a good template?

Maybe just add the following to https://github.com/actions/starter-workflows/blob/master/ci/docker-image.yml?

```
  run: |
    echo "$DOCKER_PASS" | docker login -u "$DOCKER_USER" --password-stdin
    docker push my-image-name:$(date +%s)
    docker logout
  env:
    DOCKER_USER: ${{ secrets.DOCKER_USER }}
    DOCKER_PASS: ${{ secrets.DOCKER_PASS }}
```

Note #96.

pascalandy:
This is where we disagree. Personally, I want to see a 'advance' template as well so I can see what are my options. Many of us a power users over here right? 

> but as a starter workflow is likely to cause confusion and has an awful lot of information that isn't relevant to many users. It looks like a great blog post or how-to article somewhere else though.

pascalandy:
For everyone follow this, you might be interested by our [template over here](https://github.com/firepress-org/ghostfire/tree/master/.github/workflows). It's the core one we use for all our github actions project.

Issue #57, opened 2019-08-20
[node.js] npm publish-gpr failed with error 500


zachguo:
To reproduce, run the publish-gpr workflow on a private repo with `gpr` successfully configured.
```
...
npm ERR! code E500
npm ERR! 500 Internal Server Error - PUT https://npm.pkg.github.com/@my-scope%2fmy-repo - Internal Server Error
```

0383988090:
Lam s de nhan được co đe

wrg1977:
https://twitter.com/wrg99

okuryu:
FYI: https://github.com/actions/setup-node/issues/53 may be helpful.

zachguo:
The issue is now gone, probably Github has fixed it.

PR #85, opened 2019-08-30
fixed npm publishing for NPM

codeofsumit:
As GPR only supports scoped packages, we have to put `@username` into the `name` property in the `package.json`. This is a scoped package for npm too, which then needs `--access public` to be properly published.

Or what other way do we have to define the package as scoped for GPR?

### DIFF
--- a/ci/npm-publish.yml
+++ b/ci/npm-publish.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           node-version: 12
           registry-url: https://registry.npmjs.org/
-      - run: npm publish
+      - run: npm publish --access public
         env:
           NODE_AUTH_TOKEN: ${{secrets.npm_token}}
 


zachguo:
@codeofsumit Does GPR work for you? Have you encountered any issue like #57 ?

codeofsumit:
I have not encountered that particular issue, no. GPR worked for me.

andymckay:
Thank you @codeofsumit, as I understand this would make the package be scoped with public visibility https://docs.npmjs.com/creating-and-publishing-scoped-public-packages#publishing-scoped-public-packages by default. 

Publishing packages to NPM with public scoping by default feels like it might lead to someone publishing something as public when they didn't expect that and we should do anything we can to avoid that. I think we should stick to the out of the box experience for NPM.

PR #87, opened 2019-09-01
Improve npm publish action

kidonng:
- ~~`setup-node@v1` doesn't support registry auth (GPR etc. needs this), use master branch instead~~
- Limit publish action to tags only (https://github.com/actions/starter-workflows/pull/56#issuecomment-523212856)
- Use current Node LTS version

### DIFF
--- a/ci/npm-publish.yml
+++ b/ci/npm-publish.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: actions/setup-node@master
+      - uses: actions/setup-node@v1
         with:
           node-version: 10
           registry-url: https://npm.pkg.github.com/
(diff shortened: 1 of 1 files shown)


denlb:
i think this should be enough 
https://github.com/actions/starter-workflows/pull/56/files#diff-4d526fd60b0621f46f44ad6e3befe468R5

kidonng:
> I think this should be enough for publishing only on release
> [actions/starter-workflows/pull/56/files#diff-4d526fd60b0621f46f44ad6e3befe468R5](https://github.com/actions/starter-workflows/pull/56/files#diff-4d526fd60b0621f46f44ad6e3befe468R5)

I think when I made this PR GitHub doesn't have the `release` event.

> About LTS, you can just use [matrix](https://github.com/actions/starter-workflows/blob/master/ci/node.js.yml#L12) [](/actions/starter-workflows/blob/HEAD@{2019-09-22T12:18:29Z}/ci/node.js.yml#L12) on the build stage if you want to, but i would rather use newest Node.js for publishing

LTS version is mostly supported and used, thus I would like to see it set to default. But Node.js 12 is going to be LTS exact one month later so I guess keeping this is fine 😉

PR #88, opened 2019-09-01
Initial commit of PHP template

### LINKED ISSUE #15: Create a php template
It's a bit sad not to see PHP included in this first release.

christeredvartsen:
Added simple PHP starter template along with properties file.

There doesn't seem to be any `actions/setup-php` yet (at least not public), so I skipped a build matrix using several PHP versions.

Resolves #15.

### DIFF
--- a/ci/php.yml
+++ b/ci/php.yml
@@ -1,0 +1,20 @@
+name: PHP Composer CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Validate composer.json and composer.lock
+      run: composer validate
+
+    - name: Install dependencies
+      run: composer install --prefer-dist --no-progress --no-suggest
+
+    - name: Run test suite (requires test script to be defined in composer.json)
+      run: composer run-script test
--- a/ci/php.yml
+++ b/ci/php.yml
@@ -1,4 +1,4 @@
-name: PHP Composer CI
+name: PHP Composer
 
 on: [push]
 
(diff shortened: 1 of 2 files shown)


andymckay:
Thank you @christeredvartsen this looks good to me. Will `composer run-script test` always work? It looks like based on your comment that a user will have to define this. I wonder if it would be worth adding some comments in the YAML file explaining to a user what they'd have to do?

linuxjuggler:
Actually, there is this a GitHub action which can be used to setup PHP https://github.com/shivammathur/setup-php and I think it would be nice if it become part of the repository.

christeredvartsen:
> Will `composer run-script test` always work? It looks like based on your comment that a user will have to define this. I wonder if it would be worth adding some comments in the YAML file explaining to a user what they'd have to do?

Users will have to specify this script themselves, yes. Composer has no support for `--if-present`, like for instance `npm`, so running the script will fail if it is not present. It makes sense to have a comment regarding this.

To add the script one can for instance add the following to the `scripts` object in `composer.json`:

```json
"test": "vendor/bin/phpunit"
```

More docs regarding this is available here: https://getcomposer.org/doc/articles/scripts.md

andymckay:
Could we add some comments about that to the workflow please @christeredvartsen and then let's merge this and get this added? 🙇

andymckay:
> Actually, there is this a GitHub action which can be used to setup PHP https://github.com/shivammathur/setup-php and I think it would be nice if it become part of the repository.

Cool thanks @linuxjuggler for that pointer. We'll bear that in mind, at the moment with comments I'd like to hopefully merge this one so we can get some support for our PHP users.

ci/php.yml
```diff
@@ -0,0 +1,20 @@
+name: PHP Composer CI
```
andymckay: Let's remove the word `CI` based on the updated guidelines we are trying to merge in #156. It probably makes sense to keep using the word `Composer` to show that we are using composer.

ci/php.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Validate composer.json and composer.lock
+      run: composer validate
+
+    - name: Install dependencies
+      run: composer install --prefer-dist --no-progress --no-suggest
+
+    - name: Run test suite (requires test script to be defined in composer.json)
```
andymckay: Could you add in a comment here please?

christeredvartsen:
You two comments is resolved.

andymckay:
Thank you @christeredvartsen

PR #117, opened 2019-09-12
Fix build command spelling mistake.

nshCore:
change gradlew to gradle

### DIFF
--- a/ci/android.yml
+++ b/ci/android.yml
@@ -14,4 +14,4 @@ jobs:
       with:
         java-version: 1.8
     - name: Build with Gradle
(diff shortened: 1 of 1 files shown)


andymckay:
Thanks @nshCore, this uses the https://github.com/actions/setup-java step and the gradle wrapper  `gradlew` seems to work and be the recommended approach. I'm not a Java expert so I don't know the details, could you maybe explain why we should use `gradle` instead of `gradlew`? 

We could also open an issue on the `actions/setup-java` repo to pull in some people who might know more.

Issue #21, opened 2019-08-11
Swift template


DevAndArtist:
Hopefully we'll see also more Swift related templates (and package registry support) in the next 90+ days. Keep it up GitHub. ❤️

DevAndArtist:
Not setting the bar, but it would be cool if we had multiple templates:
- macOS + Xcode
- macOS (no Xcode)
- Linux

PR #118, opened 2019-09-12
Xcode CI

maustinstar:
By default, it works with any Swift or Objective-C iOS project. Users can quickly change the `matrix.platform` property for non-iOS and Multiplatform apps.

Additionally, this works out of the box without users specifying a project name. The **Set Default Scheme** conveniently looks for a valid scheme within the project to build. This default scheme can easily be overridden by changing `matrix.scheme`.

It is set to build and test with iPhone Xs and iPhone Xs Max simulators. Again, this can quickly be changed by the `matrix.device` parameter.

Overview of the yaml:
**Matrix** lets users specify build schemes and simulator devices.
**Set Default Scheme** step looks for first valid scheme, so the CI "just works" without users changing the file.
**Build** step builds for scheme and simulator.
**Test** step tests the scheme on simulator


ci/xcode
```diff
@@ -0,0 +1,50 @@
...
+    
+    - uses: actions/checkout@v1
+    
+      # Runs if matrix.scheme is set to default
+      # Finds the first listed scheme
+    - name: Set Default Scheme
+      if: matrix.scheme == 'default'
+      run: |
+        scheme_list=$(xcodebuild -list -json | tr -d "\n")
+        default=$(echo $scheme_list | ruby -e "require 'json'; puts JSON.parse(STDIN.gets)['project']['targets'][0]")
```
ericlewis: this will not work with straight SwiftPM projects, the JSON changes to something like: 
```JSON
{"workspace" : {"name" : "THOTP", "schemes" : ["THOTP"]}}
```

maustinstar: From my understanding, this should still work for Xcode applications other than just a SwiftPM project?

Should that be incorporated? Or does that belong in a separate workflow?

I don't want this workflow to get too incomprehensible.

There is a PR open specifically for SwiftPM: #21

maustinstar:
@andymckay 

In response to your first bullet, this will work for iOS, WatchOS, MacOS, and tvOS apps. So I wasn't sure what to name it.

To make a build, Xcode must compile a specific scheme for a specific platform and device. (These can be changed to other available platforms and simulators, I just threw in iOS Simulator and a couple phones as a placeholder). I used a matrix to give users options for multiple devices. This can be removed, but it severely inhibits scalability.

For simplification, I could combine build & test into one step.

Unfortunately, the first step is required for the script to be "plug and play". I could remove the first step, but the cost is that users will have to edit the script to specify their scheme before it will work (and this might not be obvious to every user).

Any feedback on how to proceed with this in mind?

andymckay:
Thanks for the feedback @maustinstar, I haven't used Xcode build at all, so your knowledge is better than mine. I agree that the point of the workflow is that it should work out of the box as much as possible so we should sacrifice complexity if we have to.

maustinstar:
@andymckay I have a couple questions still:

1) What should we do about the name? Leave it as Xcode? Or possibly break it into 4 different scripts for each platform? (Even though it'll only change one word).

2) What about the matrix? Should we remove the matrix and only let it build for one device at a time?

3) Should I combine steps 2 and 3 (build and test) into one step?

andymckay:
> What should we do about the name? Leave it as Xcode? Or possibly break it into 4 different scripts for each platform? (Even though it'll only change one word).

Let's leave it as xcode and see how it looks.

 > What about the matrix? Should we remove the matrix and only let it build for one device at a time?

I would recommend yes, let's get it as simple as possible. Users will be able to add in the matrix and expand it.

> Should I combine steps 2 and 3 (build and test) into one step?

Yes please.

ci/xcode
jfb8618: Check out the PR i just posted this week, my approach is a lot simpler than yours but includes support to force Xcode into version 11.1 (default on the github runner is to use 10.3), which might be why Xcode SwiftPM doesn't work.  I did not use matrix schemes or ruby scripts so my approach is a little less flexible but maybe easier to incorporate for some people. 

My current project loads Lottie via SwiftPM and it builds on my GitHub Action CI well.

jfb8618:
I proposed an alternate, simplified xcode 11 / iOS CI pipeline that I've been using to good effect in my iOS project. #172 
Let me know what you guys think.

jai:
@andymckay I've addressed all your comments (was looking for an iOS starter workflow and couldn't find one) - https://github.com/actions/starter-workflows/pull/460

This scopes it to iOS though because unless I force someone to edit the workflow before use, there's no way to make it plug-and-play for all platforms. Changing it from iOS to AppleTV etc 

To make it work for AppleTV or WatchOS one has to modify the `platform` to either `tvOS Simulator` or `WatchOS Simulator` - this is true for any `xcodebuild`-based workflow (https://apple.co/2VjdF85).

Anyways just trying to call attention to my PR :)

/end_shameless_plug

ci/xcode
ashtont8: > this will not work with straight SwiftPM projects, the JSON changes to something like: 
> 
> ```JSON
> 
> {"workspace" : {"name" : "THOTP", "schemes" : ["THOTP"]}}
> 
> ```

ashtont8 (approved):

FREELANCEGITIT:
Thanks

PR #129, opened 2019-09-17
don't want to publish pre-reviewed code

alex-at-cascade:
As written, this rule will fire when a PR is opened *against* master.  But that would mean *pre-reviewed code* would be published, which seems incorrect.


andymckay:
Thank you @alex-at-cascade 🙇 I think we got this in #56.

PR #136, opened 2019-09-20
Added support for PHP and Laravel

svikramjeet:
Added ci/php.yml for php support :)

### DIFF
--- a/ci/Laravel.yml
+++ b/ci/Laravel.yml
@@ -1,0 +1,36 @@
+name: Laravel Git action
+on: [push]
+jobs:
+  run:
+    name: Run
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      max-parallel: 3
+      matrix:
+        operating-system: [ubuntu-latest]
+        php-versions: ['7.1','7.2','7.3']
+    services:
+      postgres:
+        image: postgres:11
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_PASSWORD: supersecret
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master  
+    - name: Testing PHP version
+      run: php -v
+    - name: Testing Composer version
+      run: composer install
+    - name: Testing Extensions
+      run: php -m
+    - name: Create .env from .env.example
+      run: php -r "file_exists('.env') || copy('.env.example', '.env');"
+    - name: Generate APP key
+      run: php artisan key:generate
+    - name: Database Migration
+      run: php artisan migrate
+      env:
+        DATABASE_URL: ${{ secrets.DATABASE_URL }}
+        DB_CONNECTION: ${{ secrets.DB_CONNECTION }}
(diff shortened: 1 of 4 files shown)


svikramjeet:
@mscoutermarsh : Please review

andymckay:
👋 Thanks @svikramjeet for this pull request. I note we've got a PR for PHP in #88 and that doesn't include Laravel. Also hello @christeredvartsen for that PR.

I'm a little concerned about the complexity in the Laravel template, there's a lot to do there for the end user is there anyway to make this simpler?

At this point I'd ❤️ to see one PHP template land in Actions. But I'm not an expert on PHP and would like to see either #88 land or this one. I notice that one uses `composer run-script` where as this uses `./vendor/bin/phpunit --color=always` and I'm not sure which one is more likely to work and be supported out of the box.

svikramjeet:
Thanks @andymckay for your time, yes i can see #88 , We can use alias in composer to run scripts with composer like `composer run-script tes` and `./vendor/bin/phpunit` will work exactly same if `` alias id defined as `"test": "vendor/bin/phpunit"` in composer.json but i have added `./vendor/bin/phpunit` for simplicity

svikramjeet:
@andymckay  I have removed the `--color=always1 hope everything will be clear now, I happy to make it even simpler by removing test cases but generally i usually found it important

christeredvartsen:
In my opinion it does not make sense to use a specific PHP framework (Laravel) in a generic PHP starter workflow. I think it would make more sense to have a generic one that does not use any specific frameworks and perhaps a couple of more specific templates, for instance one for Laravel and one for Symfony.

> I notice that one uses composer run-script where as this uses ./vendor/bin/phpunit --color=always and I'm not sure which one is more likely to work and be supported out of the box.

When using `vendor/bin/phpunit` directly the project is required to have a dependency to `phpunit/phpunit`. When running `composer run-script test` the developer has to implement that script within `composer.json`, referencing whatever testing framework (s)he uses.

PHPUnit *is* the most used testing framework for PHP, but I wanted to have the starter workflow as generic as possible, which is why I used `run-script`.

svikramjeet:
@christeredvartsen : I haven't used the specific config which will only work in Laravel as laravel is one of the framework which uses composer which can be used by php only applications and other frameworks.Yes PHPUnit is common testing framework and in either case we need to run composer to install the dependencies which must be present in vendor directory after running `composer install`

christeredvartsen:
The template uses `artisan` which is the command-line interface included with Laravel.

svikramjeet:
I am talking about PHP related yaml,  yes laravel is using artisan and i have added that as well

christeredvartsen:
My bad, I didn't see you had multiple workflows, I only saw the Laravel one.

svikramjeet:
No issues , Thanks for update :)

andymckay:
Thank you @svikramjeet we've merged the PHP and Laravel ones, so although they are slightly different from yours, they are now landed so users will be able to use those. Thank you 🙇 for your help.

PR #144, opened 2019-09-25
Node.js Package improve

denlb:
Publish only if the version has changed

ありがとうございます @okuryu
https://github.com/actions/setup-node/issues/53#issuecomment-534558301

### DIFF
--- a/ci/npm-publish.yml
+++ b/ci/npm-publish.yml
@@ -28,7 +28,6 @@ jobs:
         with:
           node-version: 12
           registry-url: https://registry.npmjs.org/
-          scope: '@your-github-username'
       - run: |
           LATEST=`npm view . version`
           CURRENT=`cat package.json | jq -r .version`
--- a/ci/npm-publish.yml
+++ b/ci/npm-publish.yml
@@ -1,9 +1,12 @@
 name: Node.js Package
 
 on:
-  release:
-    types:
-      - created
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


okuryu:
Yes, I think it works fine, but I'm not sure this change is the best as a workflow of the starter-workflows.

- To publish with this change you have to use GitHub Release feature on GitHub UI
- Customers may want to trigger an event other than `releases`
- Customers may want to create git tags first via `npm version` command

There is also a useful action called [`merge-release`](https://github.com/mikeal/merge-release) created by @mikeal which automatically updates the version according to the semantic version based on your commit log, and performs git tagging and publishing packages.

I think customer requirements vary, so proposing a common workflow is difficult.

denlb:
@okuryu maybe you're right about releases, so i'll revert that but `$LATEST" != "$CURRENT` is a must-have otherwise you'll get your action failing

ci/npm-publish.yml
```diff
@@ -28,7 +28,15 @@ jobs:
         with:
           node-version: 12
           registry-url: https://registry.npmjs.org/
-      - run: npm publish
+          scope: '@your-github-username'
```
okuryu: I think the scope setting is not required for `registry.npmjs.org`.

denlb: That's true, just my preference. Reverted

andymckay:
Thank you @denlb for this pull request 🙇 

We've changed the action to only trigger on `release` when its event `types: [created]`. So hopefully this reduces occurrence of this situation. Based on that I think we can skip doing this version number check. If this doesn't fix the situation, please let us know.

PR #164, opened 2019-10-11
Fix broken link. CONTRIBUTING.md is not in the same directory as pull…

jdkelley:
…_request_template.md.


andymckay:
Thank you for spotting that @jdkelly. I don't think that would work because then you end up with a URL pointing to `https://github.com/CONTRIBUTING.md` rather than the root of the repo. Let's make it a fully qualified URL `https://github.com/actions/starter-workflows/blob/master/CONTRIBUTING.md` or even easier just remove the link.

jdkelley:
Done. Thanks @andymckay.

jdkelley:
Happy to help. 👍

PR #166, opened 2019-10-16
Add R starter workflow

jimhester:
This adds a workflow for [R packages](r-project.org). It uses a [setup-r](r-lib/actions:setup-r@master) action and two R packages
  - [remotes](remotes.r-lib.org) - to install the R package dependencies
  - [rcmdcheck](r-lib/rcmdcheck) - to run and format the output of the package checking tool used in the R ecosystem.

<details>
<summary> Contributing checklist </summary>
Thank you for sending in this pull request. Please make sure you take a look at the [contributing file](CONTRIBUTING.md). Here's a few things for you to consider in this pull request:

Some general notes:

</details>

The action used (`setup-r`) is _not_ currently part of the `actions` organization. I have written a few actions often needed for R packages at https://github.com/r-lib/actions/, which is where this `setup-r` action currently lives.

I also added a svg of the R logo, but did not try to resize it to correspond with the existing logo sizes.


jimhester:
Yes, I figured as much, in the meantime I will point the R community to my collection of actions in the `r-lib` organization, and if in the future we can change `r-lib/actions/setup-r` to `actions/setup-r`, great!

If not I think that is OK as well.

nealrichardson:
@chrispat great chatting with you today at GitHub Universe. Looking forward to seeing if we can find an owner for R within GitHub so we can move this forward.

ethomson:
👋  Hi @jimhester - sorry about the delay here.  We've changed the requirements a bit here, and we're now able to support starter workflow templates that use third-party actions.

In particular:

> This workflow must only use actions that are produced by GitHub, in the actions organization, or
> 
> This workflow must only use actions that are produced by the language or ecosystem that the workflow supports. These actions must be published to the GitHub Marketplace. Workflows using these actions must reference the action using the full 40 character hash of the action's commit instead of a tag. Additionally, workflows must include the following comment at the top of the workflow file:
>
>   
>     This workflow uses actions that are not certified by GitHub.
>     They are provided by a third-party and are governed by
>     separate terms of service, privacy policy, and support
>     documentation.

In addition - we've updated the checkout action since you created this PR.  Could you update it to `actions/checkout@v2`?

Finally, we've changed the triggers that we use by default for CI templates.  Could you update the triggers to be:

```
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
```

Thanks so much!

jimhester:
@ethomson I tweaked the workflow as suggested, adding the comment, changing the triggers, pinning the commit for the setup-r action and bumping the checkout version to v2.

I also added a link in the comment to where users can find more workflow examples for R packages, hope that is OK.

Let me know if there is anything else you might need, thanks for circling back to this!

ci/r.yml
```diff
@@ -0,0 +1,37 @@
...
+# See https://github.com/r-lib/actions/tree/master/examples#readme for
+# additional example workflows available for the R community.
+
+name: R
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
andymckay, line 15:
```suggestion
    branches: [ $default-branch ]
  pull_request:
    branches: [ $default-branch ]
```

ci/r.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  build:
+    runs-on: macOS-latest
```
andymckay, line 19: I'm curious that its running `macOS-latest`, could this run on `ubuntu-latest` instead? The cost for a user of running on MacOS is significantly higher than running on Ubuntu, so we'd like to default to that if we can to prevent any surprises.

PR #172, opened 2019-10-23
Example iOS.yml for GitHub Actions ready to go!

jfb8618:
This is an example iOS Xcode CI build that I've been running successfully in my iOS 13 / Xcode 11 project with the latest GitHub Actions beta. MacOS latest resolves to 10.14 which has Xcode 11.1 installed in it. 

I included some comments to show various options you can use to have the server tell you what Xcodes it has installed / configure your build. It is setup to run tests across 4 different iOS devices on one build. 

The script includes running your build and then running unit tests inside this build. Those would need to be setup in Xcode prior to running a CI, which is pretty standard these days. 

I also included a dependencies line that updates Cocoa pods since that is commonplace too. 

I did not see any issues with code signing running like this, and have run CI a lot in the past week or two on my own private repo. 

Hope it helps someone because it took me a lot of random Google searching to sleuth and put all of this together.  I have plans to expand this into a more advanced version too since I'm really loving GitHub Actions so far. ❤️

Let me know if you need more info anywhere!

### DIFF
--- a/ci/ios.yml
+++ b/ci/ios.yml
@@ -1,0 +1,53 @@
+name: iOS CI Example
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: macOS-latest
+    
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: CocoaPod Install
+        run: pod install
+      
+      #- name: Testing....display available Xcodes on the Server
+      #  run: ls /Applications | grep Xcode
+      
+      - name: Force xcode 11 
+        run: sudo xcode-select -switch /Applications/Xcode_11.1.app
+      
+# - how to run build/test all in one line for starters
+#      - name: iPhone 11 Test
+#        run:  xcodebuild test -scheme affirm-me-test -workspace affirm-me.xcworkspace -destination 'platform=iOS Simulator,name=iPhone 11,OS=13.1'
+#      - name: iPad Air Test
+#        run:  xcodebuild test -scheme affirm-me-test -workspace affirm-me.xcworkspace -destination 'platform=iOS Simulator,name=iPad Air (3rd generation),OS=13.1'
+      
+# Split build into build-only and test-only, is faster than the above method      
+      - name: iOS build setup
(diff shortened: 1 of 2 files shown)


jeroenleenarts:
There's one thing in this script. The checkout command is tied to master. So when I push to any random branch, it builds master.

It would be better, I think, to rely on the `$GITHUB_SHA` value to checkout the changes itself and build that.
https://github.com/actions/checkout
https://help.github.com/en/github/automating-your-workflow-with-github-actions/events-that-trigger-workflows#pull-request-events-for-forked-repositories

A minor tweak, but a big improvement on you suggested work sofar. 👍

jfb8618:
@jeroenleenarts  Great catch, I didn't notice it :)  I fixed it. There's a ton of cool upgrades we can make to this kind of script but I wanted to keep this one entry-level for easy accessibility for all.

jfb8618:
Any reason you guys don't want to push this? It looks helpful. @jeroenleenarts @andymckay @mscoutermarsh @pmarsceill

lionello:
I've tested this and it works! But I had to make a few changes:
* the `,OS=13.1` for the destinations is optional and in fact can be removed or it breaks whenever the simulators get updated
* the correct casing should be `runs-on: macos-latest` (all lowercase)
* the `-workspace` flag appears to be optional (worked fine for me with only the `-scheme`)

jai:
> * the `-workspace` flag appears to be optional (worked fine for me with only the `-scheme`)

@lionello on above, are you building a Project or a Workspace? I'm trying to do it without the `-workspace` parameter but don't seem to be able to. Also, as per `xcodebuild` docs located at https://www.manpagez.com/man/1/xcodebuild/ -
> To build an Xcode workspace, you must pass both the -workspace and
     -scheme options to define the build.

Anyways here's the current copy (other items above have all been taken care of and I made it automatically select the device as well) - I don't think I can touch someone else's PR so I opened a new one: https://github.com/actions/starter-workflows/pull/460

PR #179, opened 2019-11-02
Add Docker push to GitHub Package Registry

josh:
I'm very excited about Docker support in [GitHub Package Registry](https://help.github.com/en/github/managing-packages-with-github-package-registry/configuring-docker-for-use-with-github-package-registry).

I wanted to create a starter workflow for people already using and familiar with Docker Hub's [automated builds](https://docs.docker.com/docker-hub/builds/) and [automated testing](https://docs.docker.com/docker-hub/builds/automated-testing/) workflow.

This PR replicates 3 of the core behaviors.

1. Run docker compose based test suite. Known as ["Autotest"](https://docs.docker.com/docker-hub/builds/automated-testing/) on Docker Hub.
2. Build and publish `master` to `latest` tag. While `master` is a git convention, `latest` is the default tag that's pulled by default.
3. Also build and publish tags to corresponding names.

Each of these behaviors is easily modified or disabled if someone only wants to publish releases rather than any pushes to `master`.

Would love feedback on some of the naming conventions I used in the workflow.

##

Paired with @myobie
Advice from @mattyoho and @pmarsceill

### DIFF
--- a/ci/docker-push.yml
+++ b/ci/docker-push.yml
@@ -1,0 +1,73 @@
+name: Docker
+
+on:
+  push:
+    # Publish `master` as Docker `latest` image.
+    branches:
+      - master
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+  # Run tests for any PRs.
+  pull_request:
+
+env:
+  # TODO: Change variable to your image's name.
+  IMAGE_NAME: image
+
+jobs:
+  # Run tests.
+  # See also https://docs.docker.com/docker-hub/builds/automated-testing/
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Run tests
+        run: |
+          if [ -f docker-compose.test.yml ]; then
+            docker-compose --file docker-compose.test.yml build
+            docker-compose --file docker-compose.test.yml run sut
+          else
+            docker build . --file Dockerfile
+          fi
+
+  # Push image to GitHub Package Registry.
+  # See also https://docs.docker.com/docker-hub/builds/
+  push:
+    # Ensure test job passes before pushing image.
+    needs: test
+
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Build image
+        run: docker build . --file Dockerfile --tag image
+
+      - name: Log into registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
+
+      - name: Push image
+        run: |
+          IMAGE_ID=docker.pkg.github.com/${{ github.repository }}/$IMAGE_NAME
+
+          # Strip git ref prefix from version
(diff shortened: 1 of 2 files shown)


ci/docker-push.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Login
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
+
+      - name: Push
+        run: |
+          IMAGE_ID=docker.pkg.github.com/${{ github.repository }}/$IMAGE_NAME
+
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
```
josh: I was hoping there would be a more straight forward way to access the "short revision" you'd get out of something like `git rev-parse --short HEAD`. Maybe supporting `${{ github.short_ref }}` could be a handy feature to upstream.

ci/docker-push.yml
```diff
@@ -0,0 +1,71 @@
...
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Build
+        run: docker build . --file Dockerfile --tag image
+
+      - name: Login
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
```
josh: Security check: Is this the best way to pass the auth token?

I was hoping to also silence this warning.

```
WARNING! Your password will be stored unencrypted in /home/runner/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store
```

ci/docker-push.yml
```diff
@@ -0,0 +1,71 @@
...
+    branches:
+      - master
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+  # Run tests for any PRs.
+  pull_request:
+    types: [opened, synchronize, reopened]
```
josh: There didn't seem to be a way to accept the default types for `pull_request`. Was hoping something like `pull_request: true` would have done this.

josh:
> we maybe should use `github.repository` as the default for the env var.

GPR requires the format `OWNER/REPOSITORY/IMAGE_NAME`. We include `OWNER/REPOSITORY` (aka `github.repository`) by default. But `IMAGE_NAME` is kinda up to the author. I'm not super sure what most people are going to choose here. Maybe just the repository name repeated?

ci/docker-push.yml
anthonysterling: IIRC, just using `pull_request:` will do what you're asking here. 😸

josh: @anthonysterling thanks! That worked. Should be resolved with pushed changes.

mscoutermarsh:
FYI @clarkbw

ci/docker-push.yml
ericsciple: @josh have you considered `VERSION="$(git rev-parse --short HEAD)"`?

ericsciple: oops nm i misunderstood

PR #184, opened 2019-11-05
New Ruby Gem publish workflow to replace Ruby Push

clarkbw:

  - [x] Push to tag

Some general notes:


clarkbw:
Ok, can do!

clarkbw:
I think this is a ready to merge better alternative. 

I wanted to replace the original files but they are awkwardly named starting w/ gem instead of ruby. Happy to delete them once the docs have been updated.

PR #185, opened 2019-11-07
Add caching to Elixir workflow

iangreenleaf:
Adds caching to the Elixir workflow to speed up dependencies and compilation.

I realize this isn't strictly "as simple as is needed for the service", but my opinion is that if you want people to have the best experience getting started with Actions, you should give them a workflow that's fast out of the box. Most language builds would benefit pretty greatly from some simple caching. Adding these, for example, cut the build time for my medium-sized Elixir app from 4:30 to 1:30.


ci/elixir.yml
```diff
@@ -12,10 +12,24 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
+    - name: Cache Mix Dependencies
```
henrik, line 15: Step names should be "Sentence case" apparently. So say the PR instructions these days :)

ci/elixir.yml
```diff
@@ -12,10 +12,24 @@ jobs:
...
+        path: deps
+        key: ${{ runner.OS }}-mix-${{ hashFiles('**/mix.lock') }}
+        restore-keys: |
+          ${{ runner.OS }}-mix-
     - name: Install Dependencies
       run: |
         mix local.rebar --force
         mix local.hex --force
         mix deps.get
+    - name: Cache Build Files
```
henrik, line 27: Ditto about "Sentence case".

ci/elixir.yml
```diff
@@ -12,10 +12,24 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
+    - name: Cache Mix Dependencies
+      uses: actions/cache@v1
+      with:
+        path: deps
+        key: ${{ runner.OS }}-mix-${{ hashFiles('**/mix.lock') }}
```
henrik, line 19: https://github.com/actions/cache/blob/master/examples.md#elixir---mix has an example that is slightly different. I guess the difference is whether it only uses the top level mix.lock or not. No idea which is preferable.

henrik:
Reviewers with write access, anything special needed here? Would you accept a PR that updates it to match https://github.com/actions/cache/blob/master/examples.md#elixir---mix?

iangreenleaf:
Thanks for the review, @henrik! I had given up on this PR ever getting attention, but I'm happy to make updates if the repo maintainers indicate interest.

jclem:
The only thing I'm not sure about here is whether build caching is really going to help—this will cache based on the workflow event's SHA, but in the majority of use cases (CI), that will very rarely be the same across runs.

ci/elixir.yml
jclem: I think it's fine to play it safe here and use `**/mix.lock`.

henrik:
Thank you, @jclem!

Agree that build caching based on SHA seems odd.

I noticed this uses v1 of cache but there also seems to be a v2.

In my project I named the step “Restore cache” since the step appears before building deps, so that seemed conceptually a little clearer. But I guess it must be responsible for caching as well though I guess that happens somewhere in a post-step.

iangreenleaf:
I think I added build caching somewhat optimistically but never figured out a better way to key it than the SHA. It's fine by me if you'd like to remove that part. The deps caching is the most consistent and biggest time-saver.

jclem:
Thanks for the feedback @henrik and @iangreenleaf—going to play around with this a little bit more and should merge today.

jclem:
Alright, all set! Thank you, and apologies for my not having seen this.

henrik:
❤️

PR #187, opened 2019-11-08
Added test as last step to dotnet core template

OsirisTerje:
Many of the workflows have tests added as a last step, but the dotnet-core workflow is missing that.  This PR adds that last step for the dotnet-core workflow too. 

The test command have the  verbosity set to normal, using the 'v n'.  The default for dotnet test is minimal, but for vstest.console is normal, and in practice normal makes most sense.

### DIFF
--- a/ci/dotnet-core.yml
+++ b/ci/dotnet-core.yml
@@ -15,3 +15,5 @@ jobs:
         dotnet-version: 2.2.108
     - name: Build with dotnet
       run: dotnet build --configuration Release
+    - name: Test with dotnet
+      run: dotnet test --configuration Release  -v n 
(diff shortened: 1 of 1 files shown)


lee-11:
It may be preferable to change the existing command to `dotnet test --configuration Release` rather than add a separate command.  It's a little redundant to do both since the `test` command will build if necessary.

OsirisTerje:
@lee-11 Since this is a starter workflow, it may be easier to understand when both steps are explicit.  It also shows up as two distinct steps in the log, and that might also be easier to read. 
However, your point about the automatic build is a good one, so I added the -no-build (and --no-restore) flags to the test, to make it clear.

OsirisTerje:
@lee-11 Ping :-)  Anything else you see?

OsirisTerje:
Solves #208

ethomson:
Since we took so long to review this (sorry 😰) this ended up with merge conflicts.  I went ahead and fixed this via #427 - thanks so much.

PR #192, opened 2019-11-11
Use separate steps for install, build and test

hashtagchris:
Having separate steps makes it easier to determine what failed.

Switching from `npm ci` to `npm install` because the Windows image uses node v8.10.0 and npm v5.6.0. `ci` was introduced in npm v5.7.0.

See https://github.com/actions/virtual-environments/issues/88

### DIFF
--- a/ci/node.js.yml
+++ b/ci/node.js.yml
@@ -17,11 +17,8 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - name: install
-      run: npm install
-    - name: build
-      run: npm run build --if-present
-    - name: test
-      run: npm test
+    - run: npm install
+    - run: npm run build --if-present
+    - run: npm test
       env:
         CI: true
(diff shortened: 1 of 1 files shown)


ci/node.js.yml
```diff
@@ -17,10 +17,11 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - name: npm install, build, and test
-      run: |
-        npm ci
-        npm run build --if-present
-        npm test
+    - name: install
```
joshmgross: Think it's better here to just leave off the step names so that it uses run command for the step name

jeremyepling: Yep, leave out the `name`.

hashtagchris: ✔️

joshmgross:
Once https://github.com/actions/virtual-environments/issues/88 is resolved, we should change this back to `npm ci`

hashtagchris:
Draft PR to switch back to `npm ci` when we're ready: https://github.com/actions/starter-workflows/pull/214

PR #194, opened 2019-11-14
PHP: Use actions/cache for the Composer packages

christeredvartsen:
Add a step that uses https://github.com/actions/cache to cache installed Composer packages.


christeredvartsen:
Anyone?

christeredvartsen:
Would be great if someone could review this.

ci/php.yml
```diff
@@ -12,8 +12,18 @@ jobs:
 
     - name: Validate composer.json and composer.lock
       run: composer validate
+      
+    - name: Cache Composer packages
+      id: composer-cache
+      uses: actions/cache@v1
```
chrispat, line 18:
```suggestion
      uses: actions/cache@v2
```

PR #201, opened 2019-11-17
Create rushjs workflow

ChocPanda:
Create a new starter workflow for rushjs

Some general notes:


ChocPanda:
This is a starter workflow for people using the tool [Rushjs](https://rushjs.io/pages/intro/welcome/). The workflow is based upon the travis ci configuration that is created using the [rush init](https://rushjs.io/pages/commands/rush_init/) command and the existing nodejs workflow with an additional test step which would need to be added by the user.

andymckay:
Any luck with logo @ChocPanda

PR #222, opened 2019-11-25
Improve PHP workflow and add PHP package workflow

### LINKED ISSUE #221: Improve PHP support in starter workflows.
1) The current PHP workflow is a not of much use as it does not have any way to install extensions or customize the PHP environment. Most PHP projects have extensions as requirements which are not default.

2) Also there is no matrix workflow for testing PHP packages with different versions and supported OS.

Given these 2 reasons I want to improve the existing PHP workflow and add one with matrix workflow for PHP packages using this action [shivammathur/setup-php](https://github.com/shivammathur/setup-php). Just like other setup-* actions, I find that this provides a uniform interface to setup workflows for PHP, install extensions and customize the environment.

I know you do not accept external actions here, please consider this in case of PHP, as it does not have one tool to install extensions for all OS and it is difficult to maintain a workflow for different OS if one has to script to install extensions and keep track of them.

ad4g:
- Improve PHP workflow to add support for installing extensions as most PHP projects have extensions which are not default as requirements.
- Add a PHP package workflow to test PHP packages with different versions and supported OS.

I know you do not accept external actions here, but I have used [shivammathur/setup-php](https://github.com/shivammathur/setup-php) in both these workflows as PHP does not have one tool to install extensions for all OS and it is difficult to maintain a workflow for different OS if one has to script to install extensions and keep track of them. 

- [ ] Does not use an Action that isn't in the `actions` organization. Please See #221 

Fixes #221

### DIFF
--- a/ci/php-package.yml
+++ b/ci/php-package.yml
@@ -1,0 +1,45 @@
+name: PHP package
+
+on: [push]
+
+jobs:
+  run:
+    runs-on: ${{ matrix.os }}
+    strategy:      
+      matrix:
+        php-versions: ['7.1', '7.2', '7.3']
+        os: [ubuntu-latest, windows-latest, macOS-latest]        
+    steps:
+    - uses: actions/checkout@v1
+
+    # Setup PHP environment.
+    # Docs: https://github.com/shivammathur/setup-php/blob/master/README.md    
+    - name: Set up PHP ${{ matrix.php-versions }}
+      uses: shivammathur/setup-php@master
+      with:
+        php-version: ${{ matrix.php-versions }}
+      # extension-csv: mbstring, intl # (optional) specify the extensions you want to enable/install.
+      # ini-values-csv: short_open_tag=On # (optional) specify custom php.ini values.
+      # coverage: xdebug # (optional) specify coverage driver.
+      # pecl: false # (optional) use PECL as fallback to install extensions.
+      
+    - name: Check PHP Version
+      run: php -v
+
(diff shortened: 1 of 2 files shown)


ethomson:
I'm very excited to improve PHP support!  However:

> I know you do not accept external actions here...

That's correct, we do not.  This is a hard restriction, and it's for the safety and security of the GitHub Actions platform and ecosystem.  We'd like to see improvements to the `actions/setup-php` action to support this functionality.  I'd love to see a collaboration with https://github.com/shivammathur/setup-php - that certainly seems like a sensible path forward that leverages the expertise that @shivammathur has.

But I regret that referencing actions outside the https://github.com/actions/ organization is not something that we'll include in the getting started CI workflows at the moment.

ad4g:
@ethomson Thanks for the response.
I hope you consider this PR in the future then because in its current state the PHP Composer workflow, as I said is of limited use.

MasterOdin:
@ethomson Given the policy changes in #451, would it be possible to update this workflow then to pin `shivammathur/setup-php` to a sha and have this be merged?

shivammathur:
If the policy has changed and GitHub (@ethomson) would accept the PR with `shivammathur/setup-php`, I would love to create a new updated PR as per changes in #451.

The action is used [widely](https://github.com/search?o=desc&q=shivammathur%2Fsetup-php+path%3A%2F.github%2Fworkflows&s=indexed&type=Code) by the PHP community.

PR #241, opened 2019-12-04
Remove the last `@master` bad example

hashtagchris:
`@master` should be avoided. See https://github.com/actions/toolkit/blob/master/docs/action-versioning.md.

### DIFF
--- a/ci/google.yml
+++ b/ci/google.yml
@@ -31,10 +31,10 @@ jobs:
     steps:
 
     - name: Checkout
-      uses: actions/checkout@master
+      uses: actions/checkout@v1
 
     # Setup gcloud CLI
-    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.1.0
       with:
         version: '270.0.0'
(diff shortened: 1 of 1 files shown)


hashtagchris:
@imjohnbo @craigdbarber Does `0.1.0` have all the necessary changes? Or does a new tag need to be created?

imjohnbo:
Thanks for catching this, @hashtagchris! I think the commits to `master` since `0.1.0` _except_ [this one](https://github.com/GoogleCloudPlatform/github-actions/commit/eb89c6b0ea48535b80574aee92c6ba8c94a0c4ef) are documentation-related. 

@craigdbarber, would you like to cut a `0.2.0` release to include that commit and we'll update this starter workflow to reference it?

hashtagchris:
@imjohnbo @craigdbarber Thanks! For what it's worth, most of our actions have tags that are prefixed with `v`, e.g. `v0`, `v0.2.0`. I think this is just a convention though.

lahung1992:
Viet nam.tieng viet

imjohnbo:
Done in https://github.com/actions/starter-workflows/commit/bfe353443a94ed8fb57751d2934b5c3c2f4bd61b

Issue #220, opened 2019-11-24
Rename Gradle/Maven/Ant workflow


gindex:
Hi there 👋 I used [`ci/gradle.yml`](https://github.com/actions/starter-workflows/blob/master/ci/gradle.yml) to setup a CI pipeline for [Intellij Plugin repo](https://github.com/gindex/intellij-javadoc-html-cleaner) (Gradle + Kotlin). It helped a lot 🙂

The workflow name `Java CI` does not apply in my case as well as other non java Gradle projects. To have more inclusive name I would suggest to rename the workflow to `Gradle CI`. It would apply for Maven and Ant as well.

gindex:
Close since renamed with (#358)

PR #249, opened 2019-12-07
Rename Java CI to Gradle/Ant/Maven CI

gindex:
#220

### DIFF
--- a/ci/ant.yml
+++ b/ci/ant.yml
@@ -1,4 +1,4 @@
-name: Ant CI
+name: Java CI with Ant
 
 on: [push]
 
--- a/ci/ant.yml
+++ b/ci/ant.yml
@@ -1,4 +1,4 @@
-name: Java CI
+name: Ant CI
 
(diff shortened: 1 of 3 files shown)


ethomson:
Thanks @gindex, I understand the rationale, and I agree that we could be more descriptive about the workflows.  I wonder if instead of "Gradle CI", it might be better still to say "Java CI with Gradle", "Java CI with Ant", etc...?

gindex:
@ethomson it sounds good 👍 

I had a quick look into the repo and e.g. [Android CI](https://github.com/actions/starter-workflows/blob/master/ci/android.yml) is pretty the same Gradle build as [Java CI with Gradle](https://github.com/actions/starter-workflows/blob/master/ci/gradle.yml) (except some permissions for `gradlew`, which should be the same in the similar environment). @ethomson am I missing something?

gindex:
Close since improved with (#358)

PR #269, opened 2019-12-17
More descriptive workflow name

rachmari:
This update suggests using a more descriptive workflow name and filename. We are using this workflow template in the GitHub language guides and noticed this opportunity to clarify.

The filename seems aligned to the original use of this action, which was to publish to NPM. Now that the workflow contains both GitHub Packages and NPM registries, the filename could be more general. 

However, I noticed your recommendation in this PR template:
> The name of CI workflows should only be the name of the language or platform: for example "Go" (not "Go CI" or "Go Build")

If this is misaligned with your style preferences, I can close this out.

@clarkbw @mscoutermarsh


mscoutermarsh:
👍👍

Could you keep both files temporarily? That way we don’t break any links in production.

ci/node-publish.yml
```diff
@@ -1,4 +1,4 @@
-name: Node.js Package
+name: Publish Node.js Package
 
 on:
   release:
```
Shay-am-Collab: udostepnic tez dobrze

PR #276, opened 2019-12-20
Remove max-parallel

rachmari:
We're using this template for the new language guide for Python and @brcrista [mentioned](https://github.com/github/help-docs/pull/11924/files#r360389597) max-parallel option might a carryover from Azure. I think there are only 5 jobs that will get kicked off, so it's a little redundant and I haven't noticed other CI templates using max-parallel.


andymckay:
That makes sense, thanks @rachmari

PR #285, opened 2019-12-25
docker-push workflow: tag image with $IMAGE_NAME environment variable

ScottBrenner:
Proposing tagging the Docker image built here using the `$IMAGE_NAME` environment variable set earlier in the workflow, instead of just `image`!


ScottBrenner:
Nevermind, noticed the image is properly tagged later 😅

PR #288, opened 2019-12-30
Create google.yml

goodman-ops:

Some general notes:

### DIFF
--- a/.github/workflows/google.yml
+++ b/.github/workflows/google.yml
@@ -1,0 +1,74 @@
+# This workflow will build a docker container, publish it to Google Container Registry, and deploy it to GKE.
+#
+# To configure this workflow:
+#
+# 1. Ensure that your repository contains the necessary configuration for your Google Kubernetes Engine cluster, including deployment.yml, kustomization.yml, service.yml, etc.
+#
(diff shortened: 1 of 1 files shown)


hashtagchris:
Does this new workflow replace [ci/google.yml](https://github.com/actions/starter-workflows/blob/master/ci/google.yml)? Should that file be removed in this PR? I didn't get a chance to compare them closely.

.github/workflows/google.yml
```diff
@@ -0,0 +1,74 @@
...
+  setup-build-publish-deploy:
+    name: Setup, Build, Publish, and Deploy
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Checkout
+      uses: actions/checkout@master
+
+    # Setup gcloud CLI
+    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
```
hashtagchris: Avoid `@master`. See https://aka.ms/nomasters and https://github.com/actions/starter-workflows/pull/241

ethomson:
Thanks for the PR, @goodman-ops - but like @hashtagchris mentioned, this seems like a duplicate of the existing Google starter workflow.  If you have changes that you'd like to make, could you open a PR that updates that file?  In the meantime, I'm going to close this PR.  Thanks again!

PR #304, opened 2020-01-10
Remove hardcoded values from google.yml

remyoudemans:
The google.yml deployment workflow starter hardcoded the registry hostname as `gcr.io` even though there are other possible registry hostnames. It also hardcoded the deployment name. This puts them in env variables and adds to the comments that they need to be adjusted by the user.


ethomson:
Seems reasonable to me.

/cc @imjohnbo

remyoudemans:
@ethomson so can this change be approved?

ethomson:
@remyoudemans It means that I'm okay with it, but I'd like the original author to look at it as well.

PR #315, opened 2020-01-19
Remove Python 2.7 from default tests

endolith:
Python 2 is deprecated: https://www.python.org/doc/sunset-python-2/


Olivierdub38:
[2019_05_29 15_22 Office Lens.pdf](https://github.com/actions/starter-workflows/files/4204202/2019_05_29.15_22.Office.Lens.pdf)

PR #317, opened 2020-01-23
Django workflow

### LINKED ISSUE #212: Django example
It would be useful to have a yml example for Django. It is actually named in the readme (ci/python-django.yml), but there is nothing in the ci folder.

Bhupesh-V:
Possible FIx #212 

Some general notes:

### DIFF
--- a/ci/django.yml
+++ b/ci/django.yml
@@ -12,7 +12,7 @@ jobs:
         python-version: [3.6, 3.7, 3.8]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v1
       with:
--- a/ci/django.yml
+++ b/ci/django.yml
@@ -1,4 +1,4 @@
-name: Django Build
+name: Django CI
 
 on: [push]
 
@@ -17,12 +17,12 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - name: Install Dependencies
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
     - name: Run Tests
-      env:
+      env: #define SECRET_KEY in Secrets
         SECRET_KEY: ${{ secrets.SECRET_KEY }}
         DEBUG: False
       run: |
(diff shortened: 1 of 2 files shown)


ci/django.yml
```diff
@@ -0,0 +1,29 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install Dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+    - name: Run Tests
+      env: #define SECRET_KEY in Secrets
+        SECRET_KEY: ${{ secrets.SECRET_KEY }}
```
andymckay: Presumably this is https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-SECRET_KEY. Does this need to be set for running tests? If it isn't let's not do that.

ci/django.yml
```diff
@@ -0,0 +1,29 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
+
+    steps:
+    - uses: actions/checkout@v1
```
andymckay: Can we use v2 of checkout? https://github.com/actions/checkout

ci/django.yml
Bhupesh-V: Yes this is the SECRET_KEY, it should be present to run tests.
It is currently read form the Repository Secrets.

ci/django.yml
Bhupesh-V: Done

ci/django.yml
andymckay: I'm able to run tests without a `SECRET_KEY` being specified (although I'm just trying the basic hello world). You can also specify things like `SECRET_KEY` in `tests/test_settings.py`.

Our goal is to keep the starter workflows as simple as is possible and remove anything that isn't needed. If users need to use a key, then they can set it, but I don't think its required. If you remove this I think it will get 👍 from me 😄 🙇

Bhupesh-V: Oh, okay I actually added the config just to make sure no one should push SECRET_KEYs to github.
Anyway its a users choice.
I will remove it :+1:

andymckay:
Thank you @Bhupesh-V, it will be a bit before it shows up on the page. Thanks for your contribution 🙇

PR #320, opened 2020-01-28
Updating .NET Core SDK version to LTS

timheuer:
2.2 is no longer supported, so moving to latest SDK released for .NET core which supports and encourages LTS runtime version as well as down-level building.


timheuer:
@ethomson can we get this one is as 2.2 is out of support.

ethomson:
Thanks @timheuer.  @konradpabjan did we get this in before your starter workflow update or will we have to wait for the next?

konradpabjan:
@ethomson New Starter workflows made it out yesterday and it looks like these changes are live 🎉

PR #323, opened 2020-01-29
Updating dotnet core default

timheuer:
Adding some improvements to starter for dotnet core to get to the separated steps and building on matrix.  Not a new starter, but modifications of existing and no changes to properties.json needed.


ethomson:
Since we took so long to review this (sorry 😰) this ended up with merge conflicts.  I went ahead and fixed this via #427 - thanks so much.

PR #324, opened 2020-01-29
Create main.yml

TopSooDSoI:

Some general notes:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,17 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run a one-line script
+      run: echo Hello, world!
+    - name: Run a multi-line script
+      run: |
+        echo Add other actions to build,
+        echo test, and deploy your project.


.github/workflows/main.yml
```diff
@@ -0,0 +1,17 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build:
+
```
rdg8816: 4AEE18F83AFDEB23

.github/workflows/main.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run a one-line script
+      run: echo Hello, world!
+    - name: Run a multi-line script
+      run: |
+        echo Add other actions to build,
```
dom-d: for (var adj of [
   'Raging',
   'Happy',
   'Sad',
]) {
    for (var noun of [
'Pharmacists',
'Hippo',
'Shark',
]) {
   print('The ' + adj + ' ' + noun
);
  }
}

.github/workflows/main.yml
```diff
@@ -0,0 +1,17 @@
...
+        echo test, and deploy your project.
```
dom-d:
```suggestion
        echo test, and deploy your project.
```

PR #325, opened 2020-01-30
Fix macOS virtual machine case in Swift workflow

mkj-is:
When creating new workflow using Swift, the macOS name is not lowercase. The editor tells you the value is invalid. See for yourself on the following screenshot:

Screenshot 2020-01-30 at 14 48 11

I propose making the virtual machine lowercase so people do not get this warning during Swift package setup.

### DIFF
--- a/ci/swift.yml
+++ b/ci/swift.yml
@@ -5,7 +5,7 @@ on: [push]
 jobs:
   build:
 
-    runs-on: macOS-latest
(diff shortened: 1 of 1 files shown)


ethomson:
Thanks!  It shouldn't underline that, that's a valid name.  Despite that, I agree that this should be lowercased for consistency with the other platforms.

PR #349, opened 2020-02-17
Add --locked flag for rust action

chenrui333:
Include --locked flag to the rust action to make it the build reproducible.

### DIFF
--- a/ci/rust.yml
+++ b/ci/rust.yml
@@ -10,6 +10,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Build
-      run: cargo build --verbose
+      run: cargo build --locked --verbose
     - name: Run tests
-      run: cargo test --verbose
+      run: cargo test --locked --verbose


ethomson:
I'm not familiar enough with Rust to be able to weigh in on this.  @rylev can you help here with a review?

chenrui333:
@ethomson @rylev homebrew-core side has a better well-written issue might give you more context and impact on this change :)
https://github.com/Homebrew/homebrew-core/issues/46025

rylev:
This seems reasonable to me for builds on CI, but I'm wondering what @pietroalbini thinks about this.

pietroalbini:
I don't think adding `--locked` to the default template is a good idea, as it would make the template useless for ~50.7% of the Rust repositories here on GitHub.

The `--locked` flag will not only error out if the lockfile is not up to date, but also if `Cargo.lock` is missing. [According to the Rust project's scraping efforts](https://raw.githubusercontent.com/rust-lang/rust-repos/master/data/github.csv), of the 117465 Rust repositories with a `Cargo.toml` we found across the website only 57819 of them also have a `Cargo.lock`.

This is because the project recommends to commit the lockfile only for binaries, not for libraries. This recommendation is also encoded in `cargo new --lib`, which puts `Cargo.lock` in the `.gitignore`.

ethomson:
Thanks for the information @pietroalbini!

@chenrui333 maybe instead of adding `--locked` we should add a comment in the workflow file that instead indicates that users could add `--locked`, when this would be beneficial, and what the benefits would be?

chenrui333:
@pietroalbini I am actually surprised that most of the rust projects do not have lock file, that sounds pretty interesting. But thanks for the info.

@ethomson Sounds good, I will modify my PR to add as comment to raise the awareness. Ideally the CI build should be reproducible anywhere. :)

chenrui333:
The locked file is like `package-lock.json` in npm (yarn.lock in yarn) world or `go.sum` in go modules.

chenrui333:
looks like we indeed run into the `--locked` flag issue in the homebrew side, see discussions in Homebrew/brew#10205

PR #358, opened 2020-02-20
Update Starter Workflows

konradpabjan:
Miscellaneous improvements specified in the following issue: https://github.com/github/c2c-actions/issues/611

**New Workflows**
New workflows for `Publish Java Package with Maven` and `Publish Java Package with Gradle`. The YAML comes from the `setup-java` action that already had these examples in the README: https://github.com/actions/setup-java#setup-java

**Documentation**
Added Links to guides for workflows that have documentation: https://help.github.com/en/actions/language-and-framework-guides
This includes
- Python
- Javascript (Node)
- Java (Maven, Ant & Gradle)

**Comments**
Added lots of inline comments to the basic YAML template. This will help first time actions users understand what is going on. I didn't add comments to other workflows since the name of the step is usually self-explanatory in terms of what is being run

**Triggers**
Updated triggers to work with PR events and added branch filters. Workflow events that publish some package or do a deployment have been updated to only trigger on a release

**GitHub Packages**
Changes a few references from `GitHub Package Registry` to `GitHub Packages`

A few other minor adjustments and renames for consistency

### DIFF
--- a/ci/gradle-publish.yml
+++ b/ci/gradle-publish.yml
@@ -21,10 +21,11 @@ jobs:
         server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
         settings-path: ${{ github.workspace }} # location for the settings.xml file
 
-    - name: Build with Maven
-      run: mvn -B package --file pom.xml
+    - name: Build with Gradle
+      run: gradle build
 
-    - name: Publish to GitHub Packages Apache Maven
-      run: mvn deploy -s $GITHUB_WORKSPACE/settings.xml
+    - name: Publish to GitHub Packages
+      run: gradle publish
       env:
-        GITHUB_TOKEN: ${{ github.token }}
+        USERNAME: ${{ github.actor }}
+        PASSWORD: ${{ secrets.GITHUB_TOKEN }}
--- a/ci/gradle-publish.yml
+++ b/ci/gradle-publish.yml
@@ -1,4 +1,4 @@
-# This workflow will build a package using gradle and then publish it to GitHub packages whenever a release is created
+# This workflow will build a package using Gradle and then publish it to GitHub packages when a release is created
 # For more information see: https://github.com/actions/setup-java#publishing-using-gradle
 
 name: Gradle Package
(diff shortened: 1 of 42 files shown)


ci/gradle-publish.yml
```diff
@@ -0,0 +1,31 @@
...
+        settings-path: ${{ github.workspace }} # location for the settings.xml file
+
+    - name: Build with Gradle
+      run: gradle build
+
+    - name: Publish to GitHub Packages
+      run: gradle publish
+      env:
+        USERNAME: ${{ github.actor }}
+        PASSWORD: ${{ secrets.GITHUB_TOKEN }}
```
thomabr: Should you add a comment about these matching the values in build.gradle

ci/properties/gradle-publish.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Publish Java Package with Gradle",
```
thomabr: this is rather a large name, please make sure it fits on the card without wrapping

konradpabjan: Double checked locally, looks good. Even a little wiggle room is left

image

ci/gradle-publish.yml
konradpabjan: Added information about the variables

ci/maven-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+name: Maven Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
konradpabjan: Each workflow has `runs-on: ubuntu-latest` with the exception of `swift.yml`. I changed a few workflows so that the spacing is more consistent, but I think a the same comment on each workflow is a little redundant. It makes sense for `blank.yml` as new users who are unfamiliar with the syntax will be able to learn what it is, but for other workflows it's seems excessive

rachmari:
@konradpabjan I'm away at our team mini-summit this week but I'd love to review the comments added. I don't want to block your pull request though. If you merge this before I get a chance to review, I can always follow up with a small update if I see anything that we might want to align more closely to help docs.

ethomson:
LGTM as well.  Thanks for doing this!

konradpabjan:
Thanks for the reviews everyone! I'll merge this in and create a PR to update everything on the dotcom side!

@rachmari I didn't add a whole lot of documentation (only for some of the workflows). If there is anything that needs to be aligned or updated, we can do it later.

PR #370, opened 2020-02-27
Add Tencent Kubernetes Engine starter workflow

mingshun:

Some general notes:


ci/properties/tencent.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Build and Deploy to TKE",
+    "description": "Build a Docker container, publish it to TKE Registry, and deploy to TKE cluster.",
+    "iconName": "tencentcloud",
```
cdb, line 4: Can you add an SVG icon to the `./icons` folder with this name please?

ci/properties/tencent.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Build and Deploy to TKE",
```
cdb, line 2: Can you change the name here (and in the workflow file too) to just be the platform, such as `TKE` or `Tencent Cloud` or `Tencent Kubernetes Engine`

ci/properties/tencent.properties.json
mingshun: done

ci/properties/tencent.properties.json
mingshun: done

PR #382, opened 2020-03-04
Update python-publish.yml

### LINKED ISSUE #381: ci/python-publish.yml: Build Failing due to missing libraries
Workflow: [ci/python-publish.yml](https://github.com/actions/starter-workflows/blob/a87d9c70b65a177a35dfee03ca2d7bc49aa592f6/ci/python-publish.yml)

Job: **Build and publish**
Problem: Build fails due to missing libraries required by the package (usually mentioned in requirements.txt)

Possible Fix: Install the project-specific dependencies as well.

Vibhu-Agarwal:
Fixes #381 

**Fix made**
- Installing package-specific dependencies in **Install dependencies** job.

This would fix the problem of missing libraries error in the **Build and publish** job


oar538:
- [ ] 
Duplicate of #

ci/python-publish.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel twine
+        pip install -r requirements.txt
```
andymckay, line 25: I might be misunderstanding, but there isn't anything that requires `requirements.txt` to exist is there? As such this may fail for some users. In other workflows we've guarded this to check that the file exists:

`if [ -f requirements.txt ]; then pip install -r requirements.txt; fi`

Would that make sense here?

Vibhu-Agarwal: Oh yes, definitely. Made the change.

Vibhu-Agarwal:
@andymckay Ping!

andymckay (approved):

PR #390, opened 2020-03-09
Updated template to use azure/webapps-deploy@v2

N-Usha:
[azure/webapps-deploy](https://github.com/marketplace/actions/azure-webapp) action has been enhanced to support deployments to both Azure web apps as well as web app for containers. And we would like to promote the new version of the action as part of the starter templates.

Thanks,
Usha

### DIFF
--- a/ci/azure.yml
+++ b/ci/azure.yml
@@ -35,7 +35,7 @@ jobs:
         npm run build --if-present
         npm run test --if-present
     - name: 'Deploy to Azure WebApp'
-      uses: azure/webapps-deploy@v1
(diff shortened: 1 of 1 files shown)


konradpabjan:
Thanks for the PR @N-Usha 

Not sure when these changes will go live to dotcom (after being merged in to this repo, we have to update the workflows on the doctom side for the changes to go live)

CC @ethomson @mscoutermarsh

ethomson:
Thanks @N-Usha - happy to see a new version of the action!  😀

PR #391, opened 2020-03-10
Update c-cpp.yml

xxwzan:


ethomson:
Hello!  It looks like you may have opened this pull request accidentally, or without all the changes that you wanted to include.  As a result, I'm going to close this, but please do follow up if there are changes that you want to make here.

PR #427, opened 2020-03-17
Update .NET Core CI template

### LINKED ISSUE #208: dotnet-core.yml should use `dotnet test` if tests are to be run
Looking at the `dotnet-core.yml` Activity, the description of it on the [listing page](https://github.com/eleven-software/portal-manager/actions/new) states that it "builds and tests" a .Net Core project. And that's the Right Thing, based on what the other listed Activities do.

However, in the repo the code only does a `dotnet build` which will not run tests.  The full command should instead be `dotnet test --configuration Release` which will build *and* test the project.  Small but important difference.

ethomson:
Run a `dotnet restore` step, run tests.

Fixes #208 
Obsoletes #187 and #323 - these were 👍 but we've had some drift in the existing template and needed to resolve conflicts.

### DIFF
--- a/ci/dotnet-core.yml
+++ b/ci/dotnet-core.yml
@@ -17,5 +17,9 @@ jobs:
       uses: actions/setup-dotnet@v1
       with:
         dotnet-version: 3.1.101
-    - name: Build with dotnet
-      run: dotnet build --configuration Release
+    - name: Install dependencies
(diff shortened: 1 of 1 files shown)


ethomson:
Note: tested at https://github.com/ethomson/helloworld.net/actions/runs/57408902

PR #435, opened 2020-03-24
Updating the default runner to Windows

N-Usha:
Given that this is a .NET template, Having Windows as a default runner seems a better recommendation. Please consider the change and update the starter template. 

Thanks!

### DIFF
--- a/ci/dotnet-core.yml
+++ b/ci/dotnet-core.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
 
     steps:
     - uses: actions/checkout@v2


ethomson:
I think that this is an interesting proposal, but I don't see the advantages fully yet.  In theory, .NET Core should be cross-platform and the build should be roughly identical on both `ubuntu-latest` and `windows-latest`...  If that is the case, then keeping it Linux makes more sense, since it's the lower cost option for users and seems like a better default for that reason.  If we expect that's _not_ the case then maybe we should make this a matrix instead for more coverage.

/cc @chrispat

chrispat:
I am not sure I buy the reasoning there either.  And I agree with Ed.

andymckay:
Thanks for the pull request @N-Usha, I'm closing this based on the feedback we've gotten so far.

PR #439, opened 2020-03-27
Fix- Add Execute Permissions to gradle

josharsh:


ethomson:
This feels like a bandaid over a bigger problem: users should have the execute bit on in their repository.

josharsh:
Hi!@ethomson I did not quite understand. I have just started learning about actions. Could you please tell me where did I go wrong?

ethomson:
Hi @josharsh - sorry if I was unclear - this is more about the gradle wrapper scripts in the Git repository and not directly about GitHub Actions.

Users should `chmod +x gradlew && chmod +x gradlew.bat` and check that change into their repository.

Doing the `chmod` in the GitHub Actions run should then be unnecessary.

ethomson:
I'm going to close this since the best practice is to check in the proper permissions.  Thanks again for prompting the discussion!

PR #444, opened 2020-03-31
Adds App Service documentation references

berndverst:
This workflow assumes an App Service web app has already been created and that a user knows how to obtain the desired input values. This cannot be assumed. Therefore I have included references to the relevant documentation.

Some general notes:


bnb:
+1. This is helpful additional context that holds the hand of folks who don't already have this knowledge inherently.

berndverst:
@konradpabjan could you review this small change?

konradpabjan:
Thanks for the PR! I'll try to sync up the starter workflows sometime next week so the changes should go live sometime then

PR #448, opened 2020-04-01
Add starter workflow for ruby/setup-ruby

eregon:
This PR adds a starter workflow for [ruby/setup-ruby](https://github.com/ruby/setup-ruby), the most complete action to setup Ruby. This action is hosted in the official GitHub Ruby organization.

I'm doing this PR based on this comment by @ethomson : https://github.com/actions/setup-ruby/pull/49#issuecomment-607416625

- The name of CI workflows should only be the name of the language or platform: for example "Go" (not "Go CI" or "Go Build")

Should it just be `Ruby` here?
But then we would have two starter workflows with the title `Ruby` which seems confusing.

I followed a similar structure to `ci/node.js.yml`, and used the `matrix` form to show how to test on multiple Rubies, including alternative implementations, which [actions/setup-ruby](https://github.com/actions/setup-ruby) cannot do currently.

I think the commands are obvious and need no comment.

Some general notes:

- Does not use an Action that isn't in the `actions` organization.

Since it was suggested in https://github.com/actions/setup-ruby/pull/49#issuecomment-607416625 this was OK, I suppose it doesn't apply anymore.
The action only downloads prebuilt Ruby hosted as GitHub Releases.

### DIFF
--- a/ci/ruby.yml
+++ b/ci/ruby.yml
@@ -21,7 +21,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up Ruby
-      uses: ruby/setup-ruby@v1
+    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
+    # change this to (see https://github.com/ruby/setup-ruby#versioning):
+    # uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@ec106b438a1ff6ff109590de34ddc62c540232e0
       with:
         ruby-version: 2.6
     - name: Install dependencies
--- a/ci/ruby.yml
+++ b/ci/ruby.yml
@@ -1,3 +1,10 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will download a prebuilt Ruby version, install dependencies and run tests with Rake
+# For more information see: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby
+
 name: Ruby
 
 on:
@@ -7,18 +14,17 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  test:
 
     runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Ruby 2.6
-      uses: actions/setup-ruby@v1
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.6.x
-    - name: Build and test with Rake
-      run: |
-        gem install bundler
-        bundle install --jobs 4 --retry 3
-        bundle exec rake
+        ruby-version: 2.6
+    - name: Install dependencies
+      run: bundle install
+    - name: Run tests
+      run: bundle exec rake


eregon:
I would be interested to see a rendered version of this before merging if that's possible.

ethomson:
This adds a new starter workflow template - and I feel like if we have two very similar ruby templates then we need to provide a lot of indication about when to use which.

Instead of adding a second one, perhaps we should just make the existing ruby template(s) use the `ruby/setup-ruby` action?

/cc @chrispat to sanity check me.

eregon:
> Instead of adding a second one, perhaps we should just make the existing ruby template(s) use the `ruby/setup-ruby` action?

That would be fine be me and sounds even better.

chrispat:
Yes I think that makes sense.  I don't see a reason to have two different workflows and I also expect that we will deprecate actions/setup-ruby.

ethomson:
Cool, thanks @eregon.  I have some comments to keep things more consistent with the rest of the workflows.  Having things the same across all of them simplifies our handling of them.

If you could also put the following comment at the top of the YAML:

```
# documentation.
```

ci/properties/ruby.properties.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Ruby",
-    "description": "Build and test a Ruby project with Rake.",
+    "description": "Test a Ruby project with Rake.",
```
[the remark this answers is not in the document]
ethomson: This makes sense - but I'd like to keep it 'Build and test' for consistency with the rest of the workflows.  We may need a better way to describe how to install dependencies and test things that aren't being compiled - things like python and php are in the same category here.  In the meantime, let's keep this "build and test" and follow up to update _all_ the workflows that are affected.  I opened https://github.com/actions/starter-workflows/issues/453

ci/ruby.yml
```diff
@@ -7,18 +10,14 @@ on:
...
-  build:
+  test:
 
     runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Ruby 2.6
-      uses: actions/setup-ruby@v1
+    - uses: ruby/setup-ruby@v1
```
ethomson, line 19: Can you pin this to the `v1` hash please (using the full 40 characters)?  This is going to be a requirement for using third-party actions in a starter workflow template.

ci/ruby.yml
```diff
@@ -7,18 +10,14 @@ on:
...
 
 jobs:
-  build:
+  test:
 
     runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Ruby 2.6
```
ethomson, line 16: Can we keep the `name` for consistency?  `Set up Ruby` is fine, we needn't have the version number.

ci/ruby.yml
```diff
@@ -7,18 +10,14 @@ on:
...
       with:
-        ruby-version: 2.6.x
-    - name: Build and test with Rake
-      run: |
-        gem install bundler
-        bundle install --jobs 4 --retry 3
-        bundle exec rake
+        ruby-version: 2.6
+    - run: bundle install
+    - run: bundle exec rake
```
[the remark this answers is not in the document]
ethomson: That's fine; can you give them names though?  `Install dependencies` and `Test` (or `Test with rake`) would be great.

ci/properties/ruby.properties.json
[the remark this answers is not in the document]
eregon: OK, done.

ci/ruby.yml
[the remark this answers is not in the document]
eregon: I used `Install dependencies` and `Run tests`.

ci/ruby.yml
eregon: OK, I wanted to avoid hardcoding the version twice, that works too.

ci/ruby.yml
eregon: Using the full hash or a specific release like `v1.31.0` is not going to work though, for the same reasons none of the other starter workflows use an hardcoded version number.
That would prevent users of that action to get bug fixes (e.g. at some point GH Actions set `CLASSPATH` which broke JRuby, it's good if we can workaround rather than fail every usage of the workflow using JRuby) and new (compatible) features.
Notably new Ruby versions are considered new features in terms of versioning for `ruby/setup-ruby`.

So I think we really need `ruby/setup-ruby@v1` here.

BTW, https://github.com/actions/toolkit/blob/master/docs/action-versioning.md mentions:
```yaml
- uses: actions/javascript-action@v1 # recommended. starter workflows use this
```

eregon:
@ethomson Thanks for your feedback, I think I addressed all of it, except regarding `@v1`=>`sha` (see discussion there).

ci/ruby.yml
eregon: FWIW, I also noticed in this repo that for example `ci/aws.yml` uses `aws-actions/amazon-ecr-login@v1`.

ethomson: > Using the full hash or a specific release like v1.31.0 is not going to work though, for the same reasons none of the other starter workflows use an hardcoded version number.

Yes, I understand the tradeoff here.  Our policy is to use the full hash for third-party workflows that don't come from partners because they're immutable.

We're working on other solutions, but for now, this needs to be the full hash.  Users are welcome to adopt a different mechanism (using a tag) if they decide to trust that.

eregon: I think it's needed to trust the organization here.
From #451
> This workflow must only use actions that are produced by the language or ecosystem that the workflow supports.

I think that's a good step for security. For instance only myself, @ioquatix and `ruby` org owners  have write access to the `ruby/setup-ruby` repository.

Also, `ruby/setup-ruby` doesn't require any secret and so AFAIK there is no secret data leak possible, isn't it? Can you elaborate on your security concerns for this action?

I think almost any action maintainer you will ask will say it's unreasonable to pin to a sha like this.
In other words, using a sha is unsupported for `ruby/setup-ruby` and I would think as well for most actions out there.
Using a specific release is only supported for a short while if there is an issue with `v1`.
So `ruby/setup-ruby@v1` is the only supported form, because it's the only one that can provide bug fixes and adapt to changes in GitHub Actions (e.g., when GitHub Actions adds a MSYS2 installation on Windows).

If you absolutely want a sha here, then I see only 2 ways forward:
* I make a PR on every release of ruby/setup-ruby to update the sha here. There are currently 40 releases, so it's very frequent, about once a week. I think this will be very annoying for both me and the maintainers of this repository.
* Add a comment on top of that sha saying:
```
# this is unsupported, use ruby/setup-ruby@v1 instead
```
which of course feels silly.

Another problem, the [link to the documentation](https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby) would be often outdated, which is bad for user experience.

I can tell you Ruby users want update, and want access to new Ruby versions, and that requires updating the action, and make new builds for those new versions. Also sometimes the build flags changed, so that uses a new release of the action (and so older versions still use the older builds which is good if there is a problem).

In fact, pinning to a sha would prevent something like `ruby-version: 2.6` to get the new security release like Ruby 2.6.6. So now we have conflicting security concerns.

---

To summarize, I think GitHub needs to rethink this full sha requirement for 3rd party actions, because it simply cannot work for any maintained action. I think the existing requirements (repository inside the organization of the language, disclaimer on top of the workflow) are enough.

ioquatix: > because they're immutable.

Does this even matter? Because if you are concerned about security, all the users with access  to setup-ruby could easily inject exploits into ruby/ruby. I mean, it's not like security is any better because you pin it to a 40-character SHA sum. Or am I missing something?

ioquatix: Also, I agree, the ergonomics of using a 40-character SHA sum are horrible for all the reasons discussed above.

If GitHub is so concerned about this, why not have a "github actions" lock file which locks ALL  actions down. That's what most other package management systems do.

ethomson: > I make a PR on every release of ruby/setup-ruby to update the sha here. There are currently 40 releases, so it's very frequent, about once a week. I think this will be very annoying for both me and the maintainers of this repository.

If your action needs a weekly release cadence, then it might make sense to automate those updates.

> That would prevent users of that action to get bug fixes 

Yes, we recognize that there's a security and usability tradeoff.  Users are welcome to opt-in to rolling updates by selecting a version number, and certainly we'd be happy if they did.  But we feel that this should be opt-in, and that the default experience for starter workflows should be pinned to a particular version.  This helps minimize the risk of attacks that publish a malicious action by republishing over an existing tag - not something you would do, but account takeovers _are_ a risk, even when minimized using 2FA.  And since we don't control access to your organization, like we do to https://github.com/actions, we feel like this is a risk that we need to minimize.

If you'd like, you can add a comment in your workflow that points people to some documentation - like your action's readme - that outlines the pros and cons of pinning to a SHA vs selecting a version.  But please make it reasonably fact-based and neutral, not "# this is unsupported, use ruby/setup-ruby@v1 instead".

(Many people _do_ want to pin to a particular version, because they want to run actions that they've reviewed, so this should be a supported scenario.)

Like I said ☝️ - we realize that pinning to a SHA is not ideal and we're working on improving this.  There's product planning and engineering work involved in this.  If you'd prefer to wait until that ships, I understand completely.  But unfortunately, work on this is on hold because we're all operating at reduced capacity and we're refocusing efforts on core engineering work.

We came up with this set of guidelines as a way to allow third-party actions into the starter workflow templates to get templates like yours in more quickly.  We feel that this is a good balance of the security and usability concerns.

Again, it's definitely not perfect, and we'll improve it.

eregon: > account takeovers are a risk

Isn't it also a risk for GitHub accounts? What makes it more likely for my account vs a member of the `actions` organization? (I have 2FA enabled)

@tenderlove is part of the ruby organization and a GitHub employee, maybe we can make something so GitHub can have some added trust as to who has access to `ruby/setup-ruby` and related repositories?

> we realize that pinning to a SHA is not ideal and we're working on improving this. There's product planning and engineering work involved in this. If you'd prefer to wait until that ships, I understand completely.

**Let's wait then**, I'm not in a hurry. I think no actively maintained and non-trivial third party action would want to pin to a sha.

Do you know what GitHub plans to do to address this problem?
Will using `@v1` become acceptable for 3rd party actions?

I'll summarize my concerns with pinning to a sha for `ruby/setup-ruby` in `start-workflows`:
* Inconsistent and confusing for newcomers to GitHub Actions. They will ask themselves why there is such an unpretty hardcoded sha which goes against GitHub (and everyone really) versioning recommendations.
* Potential to make support for `ruby/setup-ruby` a nightmare, both for reporters and for maintainers: new users are likely to use that sha, and then will miss bug fixes and new features. That's first a bad experience for users, which I think GitHub doesn't want either. And then they likely file an issue to the action, where we will have to them "well it's impossible to fix bugs/add new Rubies or features since you chose to pin to a specific sha" to which they likely reply "I would never have used a hardcoded sha in the first place but that's what starter-workflows suggested and that's supposed to suggest the right thing to do". I don't think any good would come out of that, but a lot of troubles for sure.

ethomson: > Isn't it also a risk for GitHub accounts? What makes it more likely for my account vs a member of the actions organization? (I have 2FA enabled)

Definitely, it is a risk.  One of the reasons we put our actions in https://github.com/actions (and not in https://github.com/github) is so that it's an independent organization that's geared around the distribution of actions.  This lets us tighten the controls much, much more aggressively than an organization that supports day-to-day software development.

> Do you know what GitHub plans to do to address this problem?

We're working on providing actions distribution mechanisms outside of the repository itself.  So that a version number (like `1.2.3`) is immutable once it's published.  This would remove the need to pin to a sha, since now pinning to a full version number will be immutable.

> Will using @v1 become acceptable for 3rd party actions?

Just to reiterate: users can, of course, specify `@v1` for third party actions.  And you can recommend that they use that, even in the template itself.  However, automatic updates need to be opt-in.

eg, something like:

```
# To automatically use the latest version of setup-ruby,
# change this to uses: ruby/setup-ruby@v1
# For more information, see https://explanation/here.html
```

> Let's wait then, I'm not in a hurry. I think no actively maintained and non-trivial third party action would want to pin to a sha.

That's disappointing.  We have an opportunity to improve the state of ruby workflows by using your superior `setup-ruby` action.  Even if the process is not as streamlined as you'd like, it's still superior to the status quo.

eregon: > We're working on providing actions distribution mechanisms outside of the repository itself.

Pinning to a full (immutable) version would still prevent bug fixes and any update. Good for security, most likely bad for reliably running the workflow, and definitely bad for user experience when it breaks due to using a too old version.
So it'd be nicer-looking than a 40-chars sha, but otherwise identical in functionality.

I think in general actions in the `actions` organization should follow the same requirements regarding versioning. Then it would be obvious immutable versions is unsustainable.

I understand GitHub wants to mark 3rd party actions, but the top-level comments + maybe something in rendering starter workflows seems enough. Imposing an unsustainable development model based on written-in-stone versions doesn't make sense to me.

Regarding the comment in the workflow, I'll think about it.
If all users take the recommendation it would be fine (although weird and they will wonder why even have something else than `@v1` in the template, makes things more complicated and not ready-to-be-used mostly as-is), but if they don't they're going to be frustrated, and maintainers of setup-ruby are going to be annoyed.

My stance regarding versioning for this action is: "you can use a full sha or a specific release, but there are no guarantees it will keep working as soon as there is a new release of the action, so use `ruby/setup-ruby@v1` instead to get bug fixes and new Ruby versions. If you have security concerns, please discuss it with us, for instance in an issue.". So I'd write something like that as the explanation.

dentarg: > We're working on providing actions distribution mechanisms outside of the repository itself. So that a version number (like `1.2.3`) is immutable once it's published. This would remove the need to pin to a sha, since now pinning to a full version number will be immutable.

I was hoping your plan was to come up with something that would allow `@v1` to be used. As @eregon wrote, there's basically no difference between immutable version and immutable sha.

@ethomson Are you reviewing each version of a third party action before you let it appear in a starter workflow? To make sure it is safe? Or is it something else?

What about the concern @ioquatix raised about `ruby/ruby`? What about the resources `ruby/setup-ruby` downloads from the latest release of `ruby/*-dev-builder` repos? I think those could change without the `ruby/setup-ruby` version has changed.

I understand you have your policy, but it was disappointing to read that you don't have the ambition to improve the situation meaningfully.

Automobileservices4:
#448

ci/ruby.yml
ioquatix: > Pinning to a full (immutable) version would [be] good for security...

To be honest, I think you are expecting too much for just pinning to a specific commit. If an account is hijacked, there are simply too many other ways to take advantage of it. In addition, that doesn't even consider that someone like myself could be coerced (e.g. under duress by government or some other crazy scenario). So, it doesn't even take that much.

The only way to make GitHub actions truly secure is to:

- Checkout all referenced packages and build a cache of all required files.
- Checksum those files using a cryptographically secure hash.
- Invite the user to review those files (good luck with that).
- Save that cache for all future test runs.
- Provide a function to rebuild the cache, prompting users to review any differences.

Pinning to a SHA is like putting a bandaid on a broken leg and it comes across as naive. I respect everyone is trying to do their best here... but can we all agree that pinning to a specific SHA doesn't really buy any increased security?

eregon: Yes, indeed I meant as it "might be better for security in general, from some point of view". But for ruby/setup-ruby it seems currently mostly pointless regarding security because the prebuilt Rubies could change without changing the sha of the action (and they need to for *-head builds as those are automatically updated daily).

@ethomson Given that, do you think it still makes sense to enforce a full sha for ruby/setup-ruby? That means many troubles for both users and maintainers, and very little security as described in the last comments.

ethomson: > @ethomson Given that, do you think it still makes sense to enforce a full sha for ruby/setup-ruby? That means many troubles for both users and maintainers, and very little security as described in the last comments.

I feel like there's a few different conversations going on here: ergonomics, quality, and security.  I think that it makes sense to separate them.

First, the ergonomics.  Pinning to a 40 character hash is ugly.  There's no doubt about it.  We all agree, and I mention it only for completeness.

Second, the quality aspects of pinning to a hash.  Yes, using a moving tag supports automatic updates, which is nice for users in many ways.  However, I want to caution that you cannot assume that people will be doing this.  Many users explicitly _do not want this_.  It's particularly common in enterprise situations.

When GitHub Actions ships in GitHub Enterprise Server it will support working fully disconnected from the internet.  Administrators will install particular versions of actions, and there is no automatic update path there.  (Many actions will need to be modified to support this way of working, if they want to support GHES, including yours.)

But even GitHub Enterprise Cloud, and non-enterprise users of github.com want and expect to be able to pin to a sha.  This is _especially_ true of people who are doing deploys to their infrastructure, and have secrets in their workflows.  People _do_ review their actions and pin to them, and people have even created tools to [streamline the process](https://github.com/mheap/pin-github-action).

I mention this to underscore that people _will_ do this and they need to be supported.  We'll improve the ergonomics for them by moving publishing out of a git repository, so that full version numbers are immutable.  But this is a thing that people are going to do.

And finally, security.  I agree that sha pinning does not prevent _other_ vulnerabilities.  If - for example - somebody take's over the Java download site and sneaks a vulnerability into the JDK download, then pinning to a SHA doesn't really help that.  I think this is what @ioquatix meant with the bandaid on a broken leg analogy.

But: 

1. _Everybody_ who downloads the JDK will get that vulnerability.  It's not an attack unique to GitHub Actions.
2. We don't have control over that.  We trust Oracle to do their best to maintain a security posture for their downloads, just like _our_ customers trust us to do the same.

I know that you're disappointed, but the fact that other theoretical security vulnerabilities exist doesn't mean that we should ignore things that provide improvements.  We're all in this together, trying to build a secure ecosystem.  We trust the Ruby maintainers to secure your side of this; but we need to secure our side as well.

eregon: I added documentation regarding choosing a version in https://github.com/ruby/setup-ruby#versioning, and updated this PR to use a full sha with a comment on top.

I still think the full sha is very suboptimal, especially for a starter workflow (it's IMHO something newcomers to GitHub Actions shouldn't be bothered with, and it's a general consideration for GitHub Actions which should be documented but clearly here is not a great place), but let's try it.
If it doesn't work well, we'll need to reconsider, which might lead to changing the phrasing here, asking to use `@v1` here, or -- worse case -- asking to remove this starter workflow. Hopefully people using a full sha will be responsible to update it whenever needed and not report issues for old versions which nobody can fix anyway.

Regarding working offline I'm somewhat doubtful how that would work with ruby/setup-ruby, if the company has private runners they probably want to build the ruby they want in the image anyway. That's something to consider when people ask for it and probably not much before. On a related note, is there a way to detect if running on a standard runner such as GitHub's runners?

eregon:
@ethomson Ready for a final review.

ci/ruby.yml
dentarg: > I mention this to underscore that people _will_ do this and they need to be supported. We'll improve the ergonomics for them by moving publishing out of a git repository, so that full version numbers are immutable. But this is a thing that people are going to do.

@ethomson With this logic, why aren't the starter workflows in `actions/` pinning to a sha instead of `v1`?

ethomson:
Thanks @eregon - this looks great to me.  We really appreciate all the work that you've done, and we'll continue to review our processes and policies here in the hopes that we can make things easier for both you, the action authors, and our shared users.

@chrispat would you care to give this a final review?

eregon:
Let's merge this?

ethomson:
Thanks again for the work on this @eregon.  We update the templates based on this repository automatically, so this will appear as the new starter workflow for Ruby within a few days.

PR #451, opened 2020-04-03
Update pull_request_template.md

ethomson:
Update the pull request template with stylistic and policy changes.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,7 +20,7 @@ Some general notes:
 
 - [ ] This workflow must only use actions that are produced by GitHub, [in the `actions` organization](https://github.com/actions), **or**
 
-  This workflow must only use actions that are produced by the language or ecosystem that the workflow supports.  These actions must be [published to the GitHub Marketplace](https://github.com/marketplace?type=actions).  Workflows using these actions must include the following comment at the top of the workflow file:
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -5,15 +5,27 @@ Thank you for sending in this pull request. Please make sure you take a look at
...
+  Packaging workflows should run on `release` with `types: [ created ]`.
 
 Some general notes:
 
-- [ ] Does not use an Action that isn't in the `actions` organization.
-- [ ] Does not send data to any 3rd party service except for the purposes of installing dependencies.
-- [ ] Does not use a paid service or product.
+- [ ] This workflow must only use actions that are produced by GitHub, [in the `actions` organization](https://github.com/actions), **or**
+
+  This workflow must only use actions that are produced by the language or ecosystem that the workflow supports.  These actions must be [published to the GitHub Marketplace](https://github.com/marketplace?type=actions).  Workflows using these actions must reference the action using the full 40 character hash of the action's commit instead of a tag.  Additionally, workflows must include the following comment at the top of the workflow file:
```
eregon, line 23: > Workflows using these actions must reference the action using the full 40 character hash of the action's commit instead of a tag. 

I think that part is problematic, see https://github.com/actions/starter-workflows/pull/448#discussion_r404220627

ethomson: Indeed, there are tradeoffs here.  Given that we're running actions from third parties, we want to maximize the security for the ecosystem.  Right now this means 40 character hashes.  We're working on other solutions, but this is the safest mechanism at the moment.

PR #454, opened 2020-04-06
Added starter workflow for WPF .NET Core Desktop apps.

edwardskrod:

Some general notes:


ci/desktop.yml
```diff
@@ -0,0 +1,108 @@
+# This workflow will build, test and package a Wpf desktop application built on .NET Core.
```
srdjanjovcic, line 1: Capitalize WPF in uppercase.

ci/desktop.yml
```diff
@@ -0,0 +1,108 @@
...
+# 
+# $pfx_cert = Get-Content '.\SigningCertificate.pfx' -Encoding Byte
+# [System.Convert]::ToBase64String($pfx_cert) | Out-File 'SigningCertificate_Encoded.txt'
+#
+# Open the output file, SigningCertificate_Encoded.txt, and copy the string inside. Then, add the string to the repo
+# as a GitHub secret and name it "Base64_Encoded_Pfx."
+# For more information on how to configure your signing certificate for this workflow, refer
+# to https://github.com/microsoft/github-actions-for-desktop-apps#signing
+#
+# Finally, add the signing certificate password to the repo as a secret and name it "Pfx_Key."
```
srdjanjovcic, line 24: Is period integral part of the name of the secret?

ci/desktop.yml
```diff
@@ -0,0 +1,108 @@
...
+# See "Build the Windows Application Packaging project" below to see how the secret is used.
+#
+# For more information on GitHub Actions, refer to https://github.com/features/actions
+# For a complete CI/CD sample to get started with GitHub Action workflows for Desktop Applications,
+# refer to https://github.com/microsoft/github-actions-for-desktop-apps
+
+name:  Wpf .NET Core
```
srdjanjovcic, line 31: Capitalization: WPF

ci/desktop.yml
```diff
@@ -0,0 +1,108 @@
...
+      with:
+        fetch-depth: 0
+
+    # Install the .NET Core workload
+    - name: Install .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.101
+
+    # Add  MsBuild to the PATH: https://github.com/microsoft/setup-msbuild
```
srdjanjovcic, line 72: Typically, I have seen it capitalized with capital 'S': MSBuild

ci/desktop.yml
```diff
@@ -0,0 +1,108 @@
...
+    - name: Setup MSBuild.exe
+      uses: microsoft/setup-msbuild@v1.0.0
+
+    # Test
+    - name: Execute unit tests
+      run: dotnet test $env:Test_Project_Path
+
+    # Restore the Wpf application to populate the obj folder with RuntimeIdentifiers
```
srdjanjovcic, line 80: Wpf => WPF

ci/properties/desktop.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Wpf .NET Core",
+    "description": "Build, test and publish a Wpf application built on .NET Core.",
```
srdjanjovcic, line 3: Wpf => WPF

ci/desktop.yml
```diff
@@ -0,0 +1,108 @@
...
+    runs-on: windows-latest
+
+    env:
+      Appx_Bundle: Always                                       # Creates an .msixbundle/.appxbundle with the .msix/.appx files for the platform specified.
+      App_Packages_Directory: AppPackages                       # The name of the directory that will contain the application packages. The default is "AppPackages."
+      Appx_Bundle_Platforms: x86|x64                            # Include x86 and x64 platforms in the bundle. Possible values: x86|x64|ARM
+      Appx_Package_Build_Mode: StoreUpload                      # Generates the .msixupload/.appxupload file and the _Test folder for sideloading.
+      Signing_Certificate: GitHubActionsDemo.pfx                # The name of the .pfx that will be created by decoding the Base 64 secret and consumed during the packaging step.
+      Solution_Name: your-solution-name                         # Replace with your solution name, i.e. MyWpfApp.sln.
+      Test_Project_Path: your-test-project-path                 # Replace with the path to your test project, i.e. MyWpfApp.Tests\MyWpfApp.Tests.csproj.
```
ethomson, line 56: Wow, this is a _lot_ of settings.  Is all of this necessary?  For example: why would you want to set up `Test_Project_Path` here?  What if you had two test projects not one?  Wouldn't just running `dotnet test` support multiple test projects _and_ reduce the amount of configuration?

These workflows aren't necessarily meant to solve _every problem_, they're to help people get started and understand how GitHub Actions works.  The more that we can automate, the less that we require people to change, and the fewer options that we have, the better.  People can grow-up into their own scenarios once we've given them the basics.

edwardskrod: Edward, thanks for your comment!  

I can reduce the number of environment variables considerably by hardcoding a few of the settings.  That should makes things less complicated.

When it comes to the test project, unfortunately, when you run "dotnet test" at the solution level, Nuget Restore tries to restore all the projects in the solution, including the Windows Application Project.  The result is a warning and and error, right off the bat.   It doesn't affect testing.  But it doesn't look nice either and may confound our users.

dotnetTest


Notice when we specify the test project, that doesn't happen.
dotNetTestProject

What do you think?  Is that too confusing for users?

I'll work on removing some of the environment variables and address Srdjan's feedback.

Thanks again!

ethomson: Aha!  I see!  Thanks for the additional information.  Is there an example project that I can use to understand this better?  I haven't written any WPF, my .NET UI experience is limited to WinForms.

ci/wpf-dotnet-core.yml
```diff
@@ -0,0 +1,107 @@
...
+
+    # Install the .NET Core workload
+    - name: Install .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.101
+
+    # Add  MsBuild to the PATH: https://github.com/microsoft/setup-msbuild
+    - name: Setup MSBuild.exe
+      uses: microsoft/setup-msbuild@v1.0.0
```
ethomson, line 70: Can you specify the full 40 character SHA here instead of the specific version?  Since git tags are not immutable (but commit IDs are) we want to make sure that we're pinning to a particular version.

Here's the updated guidance for third-party actions in starter workflow templates:

> This workflow must only use actions that are produced by the language or ecosystem that the workflow supports. These actions must be published to the GitHub Marketplace. Workflows using these actions must reference the action using the full 40 character hash of the action's commit instead of a tag.

ci/wpf-dotnet-core.yml
```diff
@@ -0,0 +1,107 @@
+# This workflow will build, test and package a WPF desktop application built on .NET Core.
```
ethomson, line 1: Before these comments, can you add the following:

```
# documentation.
```

ci/wpf-dotnet-core.yml
edwardskrod: @ethomson 
What would that look like?   I think I found the commit:
https://github.com/microsoft/setup-msbuild/commit/2008f912f56e61277eefaac6d1888b750582aa16

Like this?
   Add  MSBuild to the PATH: https://github.com/microsoft/setup-msbuild
    - name: Setup MSBuild.exe
    - uses: microsoft/setup-msbuild@2008f912f56e61277eefaac6d1888b750582aa16

ci/desktop.yml
edwardskrod: Take a look here:    https://github.com/microsoft/github-actions-for-desktop-apps

I based this workflow on the CI pipeline I developed for this sample app.  I removed a few unnecessary things, such as versioning and simplified how the project is built and what options are available to customers out of the box.

Please let me know if you have questions.

ci/wpf-dotnet-core.yml
edwardskrod: That looks like it's right based upon this guidance:
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses

ci/desktop.yml
edwardskrod: @ethomson Hi!  I'm just checking up on this.  Is there anything else you need?

ethomson:
I'm not the .NET expert.  @chrispat or @bryanmacfarlane could you review or suggest an appropriate reviewer?  Thanks!

chrispat:
It seems fine to me.    We don't typically have a pattern of using the ENV section for inputs but I don't think there is anything wrong with it.

PR #455, opened 2020-04-07
New keywomen

271992:

- [ ] CI workflows should run on `push` to `branches: [ master ]` and `pull_request` to `branches: [ master ]`.

  Packaging workflows should run on `release` with `types: [ created ]`.

Some general notes:

  This workflow must only use actions that are produced by the language or ecosystem that the workflow supports.  These actions must be [published to the GitHub Marketplace](https://github.com/marketplace?type=actions).  Workflows using these actions must reference the action using the full 40 character hash of the action's commit instead of a tag.  Additionally, workflows must include the following comment at the top of the workflow file:
    ```
    # documentation.
    ```

### DIFF
--- a/ci/blank.yml
+++ b/ci/blank.yml
@@ -1,6 +1,6 @@
 # This is a basic workflow to help you get started with Actions
 
-name: CI
+name: homegirl
 
 # Controls when the action will run. Triggers the workflow on push or pull request
 # events but only for the master branch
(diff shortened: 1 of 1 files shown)


ethomson:
Hello!  It looks like you may have opened this pull request accidentally, or without all the changes that you wanted to include.  As a result, I'm going to close this, but please do follow up if there are changes that you want to make here.

PR #460, opened 2020-04-11
Add iOS Starter Workflow

jai:
### Starter workflow to build and test an iOS project

Runs on `macos-latest` and uses `xcodebuild` (https://apple.co/2VjdF85) to build from the command line with as few system and project dependencies as possible.

---
I was looking for a starter iOS workflow and there are these two PRs that are open:
1. https://github.com/actions/starter-workflows/pull/118 (this is for an Xcode workflow but I think that's biting off a bit more than is appropriate for a 'starter' workflow)
2. https://github.com/actions/starter-workflows/pull/172

I've addressed almost all the comments the reviewers had in each of the PRs with the exception of combining the `Build` and `Test` steps because:

1. It complicates the script and makes it hard(er) to decipher (IMO).
2. Different project structures with different testing strategies may require customization of the test target in order to run properly, and I didn't want someone starting out to have to first figure out how to decouple the `Build` and `Test` steps before being able to get up and running.
---
There aren't any dependencies on project settings, from the actual Xcode workspace/project filename to the name of the device simulator used. Only `iOS Simulator` is hardcoded (as AFAIK Apple has not changed that since the early days of Xcode).

* Scheme (https://apple.co/2VqWKAG) name is automatically selected - uses the default scheme (thanks to @maustinstar - https://github.com/actions/starter-workflows/pull/118).
* Platform is hard-coded to `iOS Simulator` which should be fine for all iOS projects.
* Device is automatically selected from any available `iPhone` device as returned by running `instruments -s` (https://apple.co/3b2rr5E)
* If the project has a workspace (a file ending in `.xcworkspace` - case insensitive) then `xcodebuild` is run with the `-workspace` argument, otherwise it builds the `.xcodeproj` file using `-project`. If neither of these exist then it'll fail, which seems to make sense to me.
---
Oh, no idea what the `.properties.json` file is for but I created it and followed the formats of the existing files 🤷🏽‍♂️


jfb8618:
Looks like a great upgrade on my PR. I just wanted to leave something simple enough for anyone to roll with. Can tell iOS isn’t as popular as the other languages so it needed some love! 

I’ll prolly absorb a few of your nice upgrades into mine now 😅

jai:
@ethomson any chance someone could take a look at this? 🙏🏽

PR #463, opened 2020-04-13
Add Scala icon

seblm:
Scala icon was blank. It is now the official Scala logo.

Original file is [scala-spiral.svg](https://www.scala-lang.org/resources/img/scala-spiral.svg).

This copy was simplified with _Save As... > optimized SVG_ feature of [Inkspace](https://www.inkscape.org) (powered by [Scour](https://github.com/scour-project/scour)).


ethomson:
Looks good to me - thanks @seblm!

PR #464, opened 2020-04-14
Add `Test` step to `go.yml` (Golang starter workflow)

jai:
Testing in Go is a first-class citizen, any starter workflow should also have a Test step.


ethomson:
Looks good to me.  @mscoutermarsh or @iheanyi know a lot more about golang than I do.  Can I get a 👍 from one of y'all?

PR #481, opened 2020-04-28
Revert "Fix image name as env var" SH

centerpro:
Reverts actions/starter-workflows#472
/tools-stashbot


centerpro:
https://github.com/dpp/simply_lift/wiki/Centerpro

PR #492, opened 2020-05-01
Fix source image name on docker tag

timdorr:
This resolves an issue created by #472


cdb:
Somehow I missed that a different PR already fixed this, so this is in production already. Thanks for the PR regardless!

PR #493, opened 2020-05-03
Ada starter workflow, icon and properties

mosteo:
This is a basic workflow to compile Ada projects with the GCC compiler, GNAT, through its project compilation tool, GPRbuild.

Some resources about Ada and GNAT: 
- https://github.com/ohenley/awesome-ada
- http://www.getadanow.com/


mosteo:
Thanks for the review, @jquorning. I've pushed your suggested fix.

mosteo:
Hello Andy,

> Thanks @mosteo and sorry for the delay. The workflow looks good, I left one suggestion to cope with `$default-branch` that we can change.

Thanks for the suggestion, I just applied it.

> I'm unclear about the SVG logo you added though. It looks like a logo from Awesome Ada and not the a logo for Ada, which would confuse users. We also need a logo to be able to redistributed under the MIT license and I can't spot anything about that logo.

Right, I did privately ask the Awesome Ada author for permission to use the logo but I see your points. 

It seems the [Hummingbird logo](http://getadanow.com/) is the one used by the GNAT compiler people. According to that page, the Ada mascot can be freely used for anything Ada-related. Also, [Wikipedia's version](https://en.wikipedia.org/wiki/Ada_(programming_language)#/media/File:Ada_Mascot_with_slogan.svg) is explicitly Public Domain. I used that one and removed the text (since I see most icons avoid it). I just committed it.

PR #494, opened 2020-05-03
Update workflow name from "WPF .NET Core" to ".NET Core Desktop"

edwardskrod:
- Update the workflow name from "WPF .NET Core" to ".NET Core Desktop"
- Update the file names to reflect this change
- Update the workflow description

After a teamwide demo on Friday, May 1st, we decided that we wanted the workflow name to reflect the wider customer target audience - WPF and Windows Forms applications (Desktop applications) that are built on .NET Core.   To do this means that we need to rename the workflow and file names to reflect this change.


ci/properties/dotnet-core-desktop.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": ".NET Core Desktop",
+    "description": "Build, test, sign and publish a desktop application built on .NET Core.",
+    "iconName": "dotnetcore",
+    "categories": ["C#", "Visual Basic", "WPF", ".NET", "WindowsForms"]
```
ethomson, line 5: These categories need to align to the languages that we match on with linguist.  @mscoutermarsh do you know this dark magic?  If not do you know who does?

edwardskrod: We'd really like to keep some version of "WindowsForms" in there since it is a kind of desktop app, similar to WPF.   But if there isn't anything that maps to it, we can let it go.  Please advise.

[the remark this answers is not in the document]
edwardskrod: @cheeb00  I tried to commit the suggestion but it looks like the change is identical to the original text.

edwardskrod: I'm going to remove "WindowsForms" to help this PR move.  We can always add it another day.

edwardskrod:
@ethomson Hey Edward!  I removed the WindowsForms property.  Hopefully everything else looks good.

vkkghjkgg:
ci/android.yml

PR #495, opened 2020-05-04
Update docker-publish.yml

ChristophDietrich:
If you change the IMAGE_NAME to something which is not "image" it will fail, because at the bottom was a hard fix "image" inside. 

Now its running.


LoicViennois:
I had the same problem and opened an issue:
https://github.com/actions/starter-workflows/issues/499
I made the same fix, works fine for me.

ChristophDietrich:
So we have now 3 reviewers and how we get the change in?

stevemar:
@cdb just an FYI that this PR (#495), and two others #503 and #500 are all attempted to fix the same bug.

PR #496, opened 2020-05-05
Add node v14

bjornstar:
Node v14 is current as of [April 21st](https://github.com/nodejs/Release)

It would make sense to test against the current version of node in addition to the LTS versions.


bjornstar:
Anyone interested in node v14?

PR #500, opened 2020-05-07
Added informative Comment

nevotheless:
People (like me) could waste a couple of minutes before figuring out that the image name needs to put there as well. Without that the action just errors out with a "image not found" message. So i guess this can be useful.


stevemar:
i think #503 is the better approach

ethomson:
Thanks @nevotheless and @stevemar - I agree that #503 requires less work for the end user, so I merged that one.  Cheers!

nevotheless:
Thanks for merging that commit, should help users in the future.

PR #502, opened 2020-05-08
Update docker-publish.yml

plockaby:
The code as it stands only works because "image" is the name of the image set on line 18. If you change the name of the image on line 18, as suggested by the "TODO" on line 17 then line 75 stops working. This should correct that.


plockaby:
I guess that this is a duplicate.

PR #503, opened 2020-05-10
fix: use the IMAGE_NAME variable

ankitpopli1891:
The PR tries to fix a behavior of the existing Github Action - "Publish Docker Image"

Current Behavior:
* changing the `IMAGE_NAME` variable breaks the build

Expected Behavior:
* changing the `IMAGE_NAME` variable should not break the build

### DIFF
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -72,5 +72,5 @@ jobs:
           echo IMAGE_ID=$IMAGE_ID
           echo VERSION=$VERSION
 
-          docker tag image $IMAGE_ID:$VERSION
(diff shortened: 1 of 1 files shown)


ci/docker-publish.yml
```diff
@@ -72,5 +72,5 @@ jobs:
           echo IMAGE_ID=$IMAGE_ID
           echo VERSION=$VERSION
 
-          docker tag image $IMAGE_ID:$VERSION
+          docker tag $IMAGE_NAME $IMAGE_ID:$VERSION
```
stevemar, line 75: Considering this is used on L58, this change makes sense to me.

ethomson:
Looks good to me.  Thanks!

PR #512, opened 2020-05-16
Add script and workflow to sync the set of starter-workflows for GHES

cschleiden:
This adds a sync script to check if a starter-workflow is compatible with GHES and a workflow to run that on every push to master. 

Steps to enable:
1. Merge to master (this will result in a failed workflow run)
2. Create `ghes` branch
3. Trigger another update to master to do the initial syncing. 

Closes https://github.com/github/c2c-actions-experience/issues/2811


script/settings.json
```diff
@@ -0,0 +1,20 @@
+{
+  "folders": [
+    "../ci",
+    "../automation"
+  ],
+  "enabledActions": [
+    "actions/checkout",
```
cdb, line 7: Do you think it'd be better to pull these "live" by iterating over the folder in the ghes-actions repo? (https://github.com/github/ghes-actions/tree/master/config/actions)

script/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    const workflowFileContent = await fs.readFile(workflowPath, "utf8");
+    const workflow = safeLoad(workflowFileContent);
+
+    for (const job of Object.keys(workflow.jobs || {}).map(
+      (k) => workflow.jobs[k]
+    )) {
+      for (const step of job.steps || []) {
+        if (!!step.uses) {
+          // Check if allowed action
+          const [actionName, _] = step.uses.split("@");
```
cdb, line 76: There may be edge cases where we only pull in v2 of an action and the workflow specifies v1, but I'm fine with skipping that for the time being, I don't see any actual live cases of us not grabbing all available version tags.

There are also cases where a direct sha is used in the workflow...I guess in that case we're hoping that the sha exists in one of the included branches/tags?

script/index.ts
```diff
@@ -0,0 +1,144 @@
...
+          if (!enabledActionsSet.has(actionName.toLowerCase())) {
+            return false;
```
cdb, line 78: What do you think about logging something here about the action that's causing this to not be included? Could be useful t to understand at a glance which other actions are used most.

cdb (approved):
Couple comments but overall this looks good to me! 

I looked over the run you did and the diff generated, and they all had expected `uses: ....` actions that aren't allowed. https://github.com/cschleiden/starter-workflows/commit/7ef3753616dc64bf49e1de92fc04d79d23f9d98e

script/settings.json
cschleiden: That could be a future improvement.

script/index.ts
cschleiden: I'm leaning towards ignoring those cases for now. Yes, can happen in theory, but as far as I'm aware we're pulling in `master` for all actions and I'd really hope any sha we reference is from the master branch.

Tomasz69-maker (changes requested):

Fernandosilvas35 (approved):

.gitignore
```diff
@@ -1 +1 @@
-script/node_modules
\ No newline at end of file
+script/**/node_modules
```
Rockyspade, line 1: update status my-branch GHES workflow #

Rockyspade (commented):
#2345 #2050 #2476

PR #514, opened 2020-05-17
Use cargo check instead of build

c-x-berger:
[`cargo check`](https://doc.rust-lang.org/cargo/commands/cargo-check.html) is usually faster than `cargo build`, but for the purposes of testing if code compiles, has the same effect.

The results of `cargo check` are cached for use by other Cargo subcommands (e.g. `cargo build`), so the run will _fail_ faster, but succeed in more-or-less the same amount of time.

### DIFF
--- a/ci/rust.yml
+++ b/ci/rust.yml
@@ -14,6 +14,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Build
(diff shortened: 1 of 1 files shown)


c-x-berger:
Wait, no, `cargo build` fails in the same amount of time.

I'm real smart.

PR #517, opened 2020-05-19
Workflow to validate starter-workflows data is setup correctly

cdb:
This adds an actions workflow and corresponding script that will validate a few basic aspects for each workflow in the repository:

- YAML file exists and parses
- JSON properties file exists and matches the expected schema
- The icon used exists

FWIW it also borrows lots of code/structure from the script @cschleiden added, thanks!!

Fixes: https://github.com/github/c2c-actions-experience/issues/2792


script/validate-data/index.ts
```diff
@@ -0,0 +1,118 @@
...
+        }
+      }
+    }
+  }
+
+  return result;
+}
+
+async function checkWorkflow(workflowPath: string, propertiesPath: string): Promise<WorkflowWithErrors> {
+  let workflowErrors: WorkflowWithErrors = {
```
cschleiden, line 68:
```suggestion
  const workflowErrors: WorkflowWithErrors = {
```

cdb: I find JS really weird with `const` being mutable, which is why I tend to use `let` in this case.  For me it shows the intent that it'll be something that changes (as we add errors into it). Fair?

PR #518, opened 2020-05-20
Don't fail GHES workflow if nothing changed

cdb:
Should fix the failure here: https://github.com/actions/starter-workflows/actions/runs/110902898

### DIFF
--- a/.github/workflows/validate-data.yaml
+++ b/.github/workflows/validate-data.yaml
@@ -1,3 +1,5 @@
+name: Validate Data
+
 on:
   push:
   pull_request:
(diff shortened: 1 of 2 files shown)


cschleiden (approved):
LGTM!

awingfield916-tech:
Bug

.github/workflows/validate-data.yaml
```diff
@@ -1,3 +1,5 @@
+name: Validate Data
+
 on:
```
awingfield916-tech, line 3: Bug

.github/workflows/validate-data.yaml
```diff
@@ -1,3 +1,5 @@
...
   push:
   pull_request:
```
awingfield916-tech, line 5: Bug

PR #520, opened 2020-05-21
Do not sync workflows without categories to GHES

cschleiden:
See https://github.com/github/c2c-actions-experience/issues/2848

### DIFF
--- a/script/sync-ghes/index.ts
+++ b/script/sync-ghes/index.ts
@@ -50,10 +50,12 @@ async function checkWorkflows(
           `${workflowId}.properties.json`
         ));
         const iconName: string | undefined = workflowProperties["iconName"];
+
+        const isBlankTemplate = workflowId === "blank";
         const partnerWorkflow = workflowProperties.categories === null;
 
         const enabled =
-          !partnerWorkflow &&
+          (isBlankTemplate || !partnerWorkflow) &&
           (await checkWorkflow(workflowFilePath, enabledActions));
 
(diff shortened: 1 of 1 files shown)


aprilh9:
Unauthorized for 561-6039484

aprilh9:
I was hacked by an iTunes back up I have contacted the FBI and the SEC cyber crime

script/sync-ghes/index.ts
```diff
@@ -33,15 +43,18 @@ async function checkWorkflows(
...
-
         const workflowId = basename(e.name, extname(e.name));
-        const workflowProperties = require(join(
+        const workflowProperties: WorkflowProperties = require(join(
           folder,
           "properties",
           `${workflowId}.properties.json`
         ));
         const iconName: string | undefined = workflowProperties["iconName"];
+        const partnerWorkflow = workflowProperties.categories === null;
```
cdb, line 53: Unfortunately right now this'll block the `blank` category - https://github.com/github/github/blob/master/config/actions_workflow_templates.json#L53 - would it be better as another configuration in your `settings.json` file do you think vs the null comparison? Or don't add the list but make an exception to the exception for `blank`?

cschleiden: Ah good catch. I was trying to avoid having another configuration that we need to keep in sync. Let me think about it, maybe special casing `blank` is the easiest solution for now.. :/

cschleiden: @cdb  k, updated

PR #540, opened 2020-06-06
Make NPM_TOKEN consistent with GITHUB_TOKEN

lukyth:
It was `npm_token` before, but for GitHub it's `GITHUB_TOKEN`. This makes them be the same, all-cap case.

### DIFF
--- a/ci/npm-publish.yml
+++ b/ci/npm-publish.yml
@@ -30,7 +30,7 @@ jobs:
       - run: npm ci
       - run: npm publish
         env:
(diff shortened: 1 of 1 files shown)


Heretique666:
> 
> 
> It was `npm_token` before, but for GitHub it's `GITHUB_TOKEN`. This makes them be the same, all-cap case.

andymckay:
Thanks @lukyth.

andymckay:
I forgot to hit the Merge button, sorry about that. Thanks for reminding me stale.

PR #545, opened 2020-06-10
Fixed issue where we were using a variable to reference the signing cert name

edwardskrod:
When I created the starter workflow, I drew heavily from a Ci.yml file that used an environment variable to store the .pfx.  However, to simplify the number of variables we used in this starter-workflow, I deleted the variable throughout the code except on one line.

Hat tip to Matteo Pagani for finding this!


windofasia:
3BMeTxyoQ9I

PR #483, opened 2020-04-28
Add Terraform

imjohnbo:
## Summary

This pull request adds a Terraform starter workflow which, upon **pull request**:
- Checks out the repository to the GitHub Actions runner
- Installs the latest version of Terraform CLI and optionally configures the Terraform CLI configuration file with a Terraform Cloud user API token
- Initializes a new or existing Terraform working directory by creating initial files, loading any remote state, downloading modules, etc.
- Checks that all Terraform configuration files adhere to a canonical format
- Generates an execution plan for Terraform

**and upon push to `master`**:
- Builds or changes infrastructure according to Terraform configuration files

### Other artifacts
- ci/properties/terraform.properties.json
- icons/terraform.svg

PR #546, opened 2020-06-11
Add D workflow

Geod24:
Description (and commit message):
```
D is a statically and strongly typed, multi-paradigm, general-purpose native programming language.
D is fully open-source and maintained by a volunteer community.
For more information, see https://dlang.org

This patch adds a minimal workflow that allows users to compile D projects that use DUB,
the official D package manager and build tool.

The workflow uses the community-maintained `setup-dlang` action,
which can install any version of `dmd` (the reference compiler),
or `ldc` (the LLVM-based, performance oriented compiler,
on any of the three platforms currently supported by Github free runners.
Support for GDC is not yet implemented.

The logo used (d.svg) originates from
https://github.com/dlang/dlang.org/blob/master/images/dlogo_2015.svg
and is available under the BSL-1.0 license
(https://github.com/dlang/dlang.org/blob/master/LICENSE.txt)

Co-Authored-By: Mathias Lang <pro.mathias.lang@gmail.com>
```

I believe this satisfies all the requirements set forth by the contributing guidelines.

It does use an external action, which was the blocker in the previous attempt (#74), but according to the current guidelines this is tolerated if the disclaimer message is present (which I added).
I also used the hash, as requested in the guideline, despite it being very user-unfriendly. The release used is here:
https://github.com/dlang-community/setup-dlang/releases/tag/v1.0.0

Official website: https://dlang.org/
Official Github organization: https://github.com/orgs/dlang
Action being used: https://github.com/dlang-community/setup-dlang

You can see that the action is in a different organization, however that is simply to simplify access management. As you can see I am a member of both organizations.

As a side note:
- it would be very nice if we could benefit from the same exemption as [Terraform](https://github.com/actions/starter-workflows/pull/483#discussion_r416882184) (which also doesn't have the disclaimer, BTW);
- The guidelines don't mention what OS the action should run on. I went with the same bet and used `ubuntu-latest`, like almost everyone else;

### DIFF
--- a/ci/d.yml
+++ b/ci/d.yml
@@ -1,0 +1,29 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+name: D
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: dlang-community/setup-dlang@7c3e57bdc1ff2d8994f00e61b3ef400e67d2d7ac
+
+    - name: 'Build & Test'
+      run: |
+        # Build the project, with its main file included, without unittests
+        dub build --compiler=$DC
+        # Build and run tests, as defined by `unittest` configuration
+        # In this mode, `mainSourceFile` is excluded and `version (unittest)` are included
+        # See https://dub.pm/package-format-json.html#configurations
+        dub test --compiler=$DC
--- a/ci/properties/d.properties.json
+++ b/ci/properties/d.properties.json
@@ -1,0 +1,6 @@
+{
+    "name": "D",
+    "description": "Build and test a D project with dub.",
+    "iconName": "d",
+    "categories": [ "D" ]
+}
--- a/icons/d.svg
+++ b/icons/d.svg
@@ -1,0 +1,216 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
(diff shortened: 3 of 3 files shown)


ci/d.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: dlang-community/setup-dlang@7c3e57bdc1ff2d8994f00e61b3ef400e67d2d7ac
```
wilzbach, line 20: Shouldn't we use master directly if this is the template for users and they have to trust setup-dlang either way? (and if they are really worried they can always lock the version).
The way I see it this might hurt the standard user as he won't be able to receive any bugfix to the build pipeline.

Geod24: Unfortunately that's the way the guidelines want it to be: https://github.com/actions/starter-workflows/blob/0170fa5239991d2141e572a1b334ea577ea4f060/.github/pull_request_template.md

Geod24:
Rebased on master. Ping @ethomson @andymckay .

Geod24:
Rebased on master. @jclem : Any chance you could review this ?

Geod24:
Rebased / retargeted. Is anything else missing ?

perwy19:
ฉันไม่รู้

ในวันที่ พ. 29 ก.ค. 2020 10:12 Mathias LANG <notifications@github.com>
เขียนว่า:

Geod24:
Rebased on `main`. Shall I try ping another contributor ? Perhaps @imjohnbo ?
It'd be great to know what is holding this PR back, considering it's carefully following all the guidelines.

ci/d.yml
Kraisornbiinthi: []()

andymckay:
So sorry for the delay @Geod24. I think this is good to merge.

Geod24:
@andymckay : Thanks!

Geod24:
@andymckay : Any update on the guidelines regarding versions ? This uses the hash, as prescribed by this repository's documentation, but most other places do not. I've listed a few examples [here](https://github.com/actions/starter-workflows/pull/74#issuecomment-642489705).

andymckay:
@Geod24 most of those are referencing actions from partners and have different guidelines. I think one or two did slip through the net and we should review those to match the new guidelines. Thank you for the heads up on that.

PR #552, opened 2020-06-16
Update php cache key to use php in its name

MasterOdin:
    ```
    # documentation.
    ```

---

Updates the cache key used in the PHP workflow to use the word `php` instead of `node` since this is a php workflow after all.


andymckay:
Thank you @MasterOdin 🙇

PR #556, opened 2020-06-23
Updates hashicorp/setup-terraform version to v1.1.0

chriswahl:
# Previous behavior
`hashicorp/setup-terraform@v1` performs a checkout of the latest Terraform _beta_ release (0.13) when no version is specified

# New behavior
`hashicorp/setup-terraform@v1.1.0` performs a checkout of the latest Terraform _stable_ release (0.12) when no version is specified


Geod24:
According to the [guidelines](https://github.com/actions/starter-workflows/blob/3c3736f59805d1e4f838182263705f44fab9cf68/.github/pull_request_template.md), this should be the 40-characters hash and come with a disclaimer.

PR #557, opened 2020-06-23
Update PHP workflow to use shivammathur/setup-php

MasterOdin:

Some general notes:

    ```
    # documentation.
    ```

-----

Continution of work done in #222 to follow new guidelines from #451

This updates the php workflow to use [shivammathur/setup-php](https://github.com/shivammathur/setup-php) action to configure PHP for a variety of versions. This is some amount of the de-facto action to do this and is used by a variety of the big PHP projects (e.g. symfony) and I would say is a good candidate for usage in the official starter file This also promotes the starter file to a similar level as the other languages (e.g. python, nodejs).

### DIFF
--- a/ci/php.yml
+++ b/ci/php.yml
@@ -29,10 +29,10 @@ jobs:
       uses: shivammathur/setup-php@dd27d4eb8158e1dd5bd52a143a7468d7a7fb302a
       with:
         php-version: ${{ matrix.php-version }}
-        # extension-csv: mbstring, intl # (optional) specify the extensions you want to enable/install.
-        # ini-values-csv: short_open_tag=On # (optional) specify custom php.ini values.
-        # coverage: xdebug # (optional) specify coverage driver.
-        # pecl: false # (optional) use PECL as fallback to install extensions.
+        # extensions: mbstring, intl # (optional) Specify the extensions you want to add or remove.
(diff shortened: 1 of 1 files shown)


ci/php.yml
```diff
@@ -10,18 +15,37 @@ jobs:
...
+    # Setup PHP environment.
+    # Docs: https://github.com/shivammathur/setup-php/blob/master/README.md    
+    - name: Set up PHP ${{ matrix.php-version }}
+      uses: shivammathur/setup-php@dd27d4eb8158e1dd5bd52a143a7468d7a7fb302a
+      with:
+        php-version: ${{ matrix.php-version }}
+        # extension-csv: mbstring, intl # (optional) specify the extensions you want to enable/install.
+        # ini-values-csv: short_open_tag=On # (optional) specify custom php.ini values.
+        # coverage: xdebug # (optional) specify coverage driver.
+        # pecl: false # (optional) use PECL as fallback to install extensions.
```
shivammathur, line 35: @MasterOdin 
Please update the optional inputs and the description
```
# extensions: mbstring, intl # (optional) Specify the extensions you want to add or remove.
# ini-values: short_open_tag=On # (optional) Specify the values you want to add to php.ini.
# coverage: xdebug # (optional) Specify the code coverage driver you want to setup.
# tools: pecl # (optional) Specify the tools you want to setup
```

MasterOdin: done

PR #571, opened 2020-06-29
elixir.yml: Specify latest Elixir/OTP versions

henrik:
Source for this being latest Elixir: https://github.com/elixir-lang/elixir/releases

There is an OTP 23.0 (https://www.erlang.org/downloads) but https://hexdocs.pm/elixir/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp indicates Elixir 1.10 isn't known to be compatible with a later OTP than 22.

### DIFF
--- a/ci/elixir.yml
+++ b/ci/elixir.yml
@@ -16,8 +16,8 @@ jobs:
     - name: Set up Elixir
       uses: actions/setup-elixir@v1
       with:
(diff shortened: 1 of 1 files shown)


jclem:
Did a quick manual test https://github.com/actions/setup-elixir/pull/34/checks?check_run_id=827719755

henrik:
Thank you, @jclem! :)

Issue #573, opened 2020-07-01
Update deprecated for actions/setup-ruby `version` to `ruby-version` on gem-push.yml


koshilife:
I got the below warning when used `starter-workflows/ci/gem-push.yml` in order to push a Ruby Gem to RubyGems.

<img width="1017" alt="Screen Shot 2020-07-01 at 23 54 47" src="https://user-images.githubusercontent.com/583336/86259445-fc738700-bbf6-11ea-8344-4d1ba2c0f6aa.png">

I think this should be fix.

For your reference that the warning code at `actions/setup-ruby` is below:
https://github.com/actions/setup-ruby/blob/9b0a0db81fa3e3966b72709c5662043856983b12/action.yml#L9

PR #574, opened 2020-07-01
Update deprecated for actions/setup-ruby  to  on gem-push.yml

koshilife:
refs #573
Update deprecated for actions/setup-ruby `version` to `ruby-version` on gem-push.yml.


koshilife:
I withdraw this PR, because the branch name is mistaken.
I created new PR instead. #575

PR #575, opened 2020-07-01
Update deprecated for actions/setup-ruby to on gem-push.yml

koshilife:
refs #573
Update deprecated for actions/setup-ruby `version` to `ruby-version` on gem-push.yml.


andymckay:
Thanks @koshilife

PR #580, opened 2020-07-08
nix: add Nix action

domenkozar:

Some general notes:

    ```
    # documentation.
    ```

---

Adds [Nix](nixos.org/) support. This allows people to quickly setup any project that uses Nix.

On top of that it integrates with a Cachix free service (for public repositories) for saving build time via binary caching.


andymckay:
We've got a starter workflow over here: https://github.com/actions/starter-workflows/pull/605 and I suspect we could do with one instead of two. A couple of things:
* This workflow won't work out of the box because of the secret, do we need to have that?
* What license is the nix logo under, this repository is under the MIT license https://github.com/actions/starter-workflows/blob/main/LICENSE

andymckay:
One more, from the pull request template 👇🏾 

This workflow must only use actions that are produced by the language or ecosystem that the workflow supports. These actions must be published to the GitHub Marketplace. Workflows using these actions must reference the action using the full 40 character hash of the action's commit instead of a tag. Additionally, workflows must include the following comment at the top of the workflow file:

# documentation.

domenkozar:
> This workflow won't work out of the box because of the secret, do we need to have that?

Technically it will work without cachix-action, but each PR will build everything from scratch.

That action makes sure that no unnecessary builds happen (sadly we can't use the cache action).

> What license is the nix logo under, this repository is under the MIT license https://github.com/actions/starter-workflows/blob/main/LICENSE

https://github.com/NixOS/nixos-homepage/commit/525ade227e23f2b54975021553eeb3c6b57c04bd

andymckay (approved):
Cool thanks @domenkozar

Newman1515 (approved):
Ya

PR #583, opened 2020-07-09
Move manual workflow template to automation category

cschleiden:
This change moves the existing manually-triggered workflow template to the `Automation` category.


koosy79 (commented):
Ok

PR #584, opened 2020-07-10
Include Bearer in GEM_HOST_API_KEY

jcansdale:
Fixes github/c2c-package-registry#2154

Ruby 2.7 now includes built in support for the `GEM_HOST_API_KEY` env var. When set, this is used in preference to the `--KEY` / `.gem/credentials` option. The `GEM_HOST_API_KEY` option _must_ include a `Bearer` prefix.

When not set, GitHub Packages will fail with:

```
Your request could not be authenticated by the GitHub Packages service. Please ensure your access token is valid and has the appropriate scopes configured.
```

https://github.com/saigkill/publican_creators/runs/853772278?check_suite_focus=true#step:5:17

### What this PR does

- Include `Bearer` in `GEM_HOST_API_KEY` env var (for compatibility with 2.7+)
- Use `${{secrets.GITHUB_TOKEN}}` instead of `${{secrets.GPR_AUTH_TOKEN}}`
- Use `${{ github.repository_owner }}` as the default `OWNER`
- Use `ruby-version` instead of `version`
  - This gets rid of the warning: `Input 'version' has been deprecated with message: The version property will not be supported after October 1, 2019. Use ruby-version instead`

### Questions

If we changed to using Ruby 2.7, the following could be removed:

```
        mkdir -p $HOME/.gem
        touch $HOME/.gem/credentials
        chmod 0600 $HOME/.gem/credentials
        printf -- "---\n:github: Bearer ${GEM_HOST_API_KEY}\n" > $HOME/.gem/credentials
```

The publish step would be simply:

```
    - name: Publish to GPR
      run: gem push --host https://rubygems.pkg.github.com/${OWNER} *.gem
      env:
        GEM_HOST_API_KEY: "Bearer ${{secrets.GITHUB_TOKEN}}"
        OWNER: ${{ github.repository_owner }}
```

Should we default to using Ruby 2.7 and/or mention this as a comment?

2. We're doing `gem build *.gemspec` as part of `Publish to GPR` and `Publish to RubyGems`. Should this be a separate step?

### DIFF
--- a/ci/gem-push.yml
+++ b/ci/gem-push.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Ruby 2.6
       uses: actions/setup-ruby@v1
       with:
-        version: 2.6.x
+        ruby-version: 2.6.x
 
     - name: Publish to GPR
       run: |
--- a/ci/gem-push.yml
+++ b/ci/gem-push.yml
@@ -27,7 +27,7 @@ jobs:
         gem build *.gemspec
         gem push --KEY github --host https://rubygems.pkg.github.com/${OWNER} *.gem
       env:
-        GEM_HOST_API_KEY: ${{secrets.GPR_AUTH_TOKEN}}
+        GEM_HOST_API_KEY: "Bearer ${{secrets.GITHUB_TOKEN}}"
         OWNER: username
 
     - name: Publish to RubyGems
@@ -39,4 +39,4 @@ jobs:
         gem build *.gemspec
         gem push *.gem
       env:
-        GEM_HOST_API_KEY: ${{secrets.RUBYGEMS_AUTH_TOKEN}}
+        GEM_HOST_API_KEY: "Bearer ${{secrets.RUBYGEMS_AUTH_TOKEN}}"
--- a/ci/gem-push.yml
+++ b/ci/gem-push.yml
@@ -28,7 +28,7 @@ jobs:
         gem push --KEY github --host https://rubygems.pkg.github.com/${OWNER} *.gem
       env:
         GEM_HOST_API_KEY: "Bearer ${{secrets.GITHUB_TOKEN}}"
-        OWNER: username
+        OWNER: ${{ github.repository_owner }}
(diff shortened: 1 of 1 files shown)


jcansdale:
@thehungrysmurf do you think we change to using Ruby 2.7? 🤔

thehungrysmurf:
@jcansdale I don't see why not, we actually setup and test [with both versions](https://github.com/github/github/blob/master/config/ruby-version#L4-L5) in dotcom, so we have high confidence there. But I'm not sure how much testing is done in Actions with different versions of ruby.

I don't have a good pulse on the ruby community, are our users using ruby 2.7 more than 2.6 now?

rngtng:
@jcansdale thanks for the hint. I updated my workflow to ruby 2.7 and it works like a charm. Had to remove `Bearer` from the secret for rubygems though.

simonini:
Hi thanks for the improvement :+1: 

With GitHub Packages everything is ok, but we had a problem with "bearer" in the RubyGems step.
We solved when we removed the bearer in the ruby gem step.

This is our commit:
https://github.com/rubynetti/amico-db/commit/eb3d2722785846184578e3b75cfd671df57bee84

Is it ok?

simonini:
@jcansdale thanks for the improvement :+1: 

@rngtng  I had the same problem and open a pull request to fix it

https://github.com/actions/starter-workflows/pull/645

jcansdale:
Hi @simonini,

Did I break publishing to RubyGems in this PR? 😊 

What does your `RUBYGEMS_AUTH_TOKEN` secret look like? Does it not contain a "Bearer" prefix? 🤔

simonini:
Hi @jcansdale  :smile: maybe yes :muscle: 

I don't know the details of RubyGems implementation but I see that without bearer it runs ok. 
Before the answer was "Access denied" :red_circle: 

You can check our action history here: https://github.com/rubynetti/amico-db/actions?query=workflow%3A%22Ruby+Gem%22

PR #590, opened 2020-07-13
Use $default-branch token

cschleiden:
This updates the existing starter workflows to use the new `$default-branch` token instead of hard-coded `master`.

### DIFF
--- a/automation/label.yml
+++ b/automation/label.yml
@@ -3,7 +3,7 @@
 #
 # To use this workflow, you will need to set up a .github/labeler.yml
 # file with configuration.  For more information, see:
-# https://github.com/actions/labeler/blob/master/README.md
+# https://github.com/actions/labeler
 
 name: Labeler
 on: [pull_request]
--- a/ci/android.yml
+++ b/ci/android.yml
@@ -2,9 +2,9 @@ name: Android CI
 
 on:
   push:
-    branches: [ master ]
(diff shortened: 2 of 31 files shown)


endolith:
This doesn't work for me.  When I commit this change, the tests no longer trigger at all.  Are we supposed to manually replace this with `master`, or is it supposed to work automatically?  I checked and it says 

> The default branch is set to `master`.

cschleiden:
@endolith 
This only works for [workflow templates(https://docs.github.com/en/actions/configuring-and-managing-workflows/sharing-workflow-templates-within-your-organization), not for workflows.

endolith:
ohhh, so the workflow is generated _from_ this, not actually a copy of it?

cschleiden:
It's a copy, but *during* the copy process the `$default-branch` variable is replaced with the default branch name of the target repository.

256SHA:
Negative

iAMnormztwizedumz:
Upon entering unlicenced right, so with just DIR only master is given, an option of numbers as per DIR explains.

bjornstar:
This is a bit confusing.

These "workflow templates" should be clearly labelled as "workflow templates" otherwise people (like me) will copy them and try to use them *as is*. Is this a repository full of workflows or workflow templates? It seems like we've had the old switcheroo pulled on us here.

PR #591, opened 2020-07-14
Add Phoenix starter workflow

jclem:
Add a Phoenix starter workflow

Phoenix is a popular web application framework for the Elixir programming language (for which we have a [starter workflow](https://github.com/actions/starter-workflows/blob/920aa0286cbaca09c2eceed9576246d6c5d38100/ci/elixir.yml)). This pull request adds a workflow similar to the Elixir workflow—the main difference is that I've added a Postgres service (Postgres is the default database for new Phoenix projects).

### Links

- [Elixir](https://elixir-lang.org)
- [Phoenix](https://www.phoenixframework.org)

### To Do

- [ ] Get SVG and understand usage rights of Phoenix logo.

---

- [ ] ~The workflow filename of publishing workflows should be the name of the language or platform, in lower case, followed by "-publish".~
- [ ] ~Packaging workflows should run on `release` with `types: [ created ]`.~

Some general notes:

    ```
    # documentation.
    ```

### DIFF
--- a/ci/phoenix.yml
+++ b/ci/phoenix.yml
@@ -18,7 +18,8 @@ jobs:
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
-          POSTGRES_DB: phoenix_test # Replace with your test database name!
+          # NOTE: Replace "phoenix_test" with your DB name from config/test.exs.
+          POSTGRES_DB: phoenix_test
         options: >-
           --health-cmd pg_isready
           --health-interval 10s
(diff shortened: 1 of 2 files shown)


jclem:
@andymckay Do you know if a PNG will work for this? I noticed that `iconName` is just a string, so I'm curious if the rendering assumes that it will be an SVG.

tonirmv (commented):
#738

Monssif1:
مرحبا

Monssif1:
انا جديد في هد الميدان

PR #592, opened 2020-07-15
Revert "Use $default-branch token"

Maxusz666:
Reverts actions/starter-workflows#590


andymckay:
I see no explanation why you are making this revert, so closing.

PR #595, opened 2020-07-18
Update .NET Core SDK version

stevedesmond-ca:
Blazor projects fail to build in 3.1.101 due to a bug in the .NET Core SDK. Updating the SDK to the latest version (3.1.301) fixes this bug.

PR #596, opened 2020-07-21
Revert "Update .NET Core SDK version"

Maxusz666:
Reverts actions/starter-workflows#595


andymckay:
This is the second PR you've attempted to revert without explanation could you please take a look at the community guidelines or explain why you are making this change?

PR #611, opened 2020-08-03
Updated google gke workflow(https://github.com/GoogleCloudPlatform/github-actions/blob/master/example-workflows/gke/.github/workflows/gke.yml)

ktoor-google:
This repository contains configuration for what users see when they click on the `Actions` tab.

It is not:

---

Thank you for sending in this pull request. Please make sure you take a look at the [contributing file](https://github.com/actions/starter-workflows/blob/main/CONTRIBUTING.md). Here's a few things for you to consider in this pull request.

Some general notes:

    ```
    # documentation.
    ```

### DIFF
--- a/ci/google.yml
+++ b/ci/google.yml
@@ -4,7 +4,7 @@
 #
 # 1. Ensure that your repository contains the necessary configuration for your Google Kubernetes Engine cluster, including deployment.yml, kustomization.yml, service.yml, etc.
 #
-# 2. Set up secrets in your workspace: GKE_PROJECT with the name of the project, GKE_EMAIL with the service account email, GKE_KEY with the Base64 encoded JSON service account key (https://github.com/GoogleCloudPlatform/github-actions/tree/docs/service-account-key/setup-gcloud#inputs).
+# 2. Set up secrets in your workspace: GKE_PROJECT with the name of the project and GKE_SA_KEY with the Base64 encoded JSON service account key (https://github.com/GoogleCloudPlatform/github-actions/tree/docs/service-account-key/setup-gcloud#inputs).
 #
 # 3. Change the values for the GKE_ZONE, GKE_CLUSTER, IMAGE, REGISTRY_HOSTNAME and DEPLOYMENT_NAME environment variables (below).
 #
--- a/ci/google.yml
+++ b/ci/google.yml
@@ -33,7 +33,7 @@ jobs:
       uses: actions/checkout@v2
 
     # Setup gcloud CLI
-    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.1.3
       with:
         service_account_key: ${{ secrets.GKE_SA_KEY }}
         project_id: ${{ secrets.GKE_PROJECT }}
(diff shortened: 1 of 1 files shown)


ci/google.yml
```diff
@@ -7,69 +7,71 @@
...
     - name: Checkout
       uses: actions/checkout@v2
 
     # Setup gcloud CLI
     - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
       with:
-        version: '270.0.0'
-        service_account_email: ${{ secrets.GKE_EMAIL }}
-        service_account_key: ${{ secrets.GKE_KEY }}
+        version: '290.0.1'
```
bharathkkb, line 38: I think we can let it default to `latest`

ci/google.yml
```diff
@@ -7,69 +7,71 @@
...
     name: Setup, Build, Publish, and Deploy
     runs-on: ubuntu-latest
-    steps:
 
+    steps:
     - name: Checkout
       uses: actions/checkout@v2
 
     # Setup gcloud CLI
     - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
```
bharathkkb, line 36: @imjohnbo Please correct me if I am wrong, but the recommendation was to pin this to a release right?

bharathkkb:
@imjohnbo PTAL

ci/google.yml
imjohnbo: That's right, it looks like the latest release is `0.1.3`, @bharathkkb 👍

ci/google.yml
```diff
@@ -7,69 +7,71 @@
 # 2. Set up secrets in your workspace: GKE_PROJECT with the name of the project, GKE_EMAIL with the service account email, GKE_KEY with the Base64 encoded JSON service account key (https://github.com/GoogleCloudPlatform/github-actions/tree/docs/service-account-key/setup-gcloud#inputs).
```
imjohnbo, line 7:
```suggestion
# 2. Set up secrets in your workspace: GKE_PROJECT with the name of the project and GKE_SA_KEY with the Base64 encoded JSON service account key (https://github.com/GoogleCloudPlatform/github-actions/tree/docs/service-account-key/setup-gcloud#inputs).
```

ci/google.yml
```diff
@@ -7,69 +7,71 @@
...
 #
 # 3. Change the values for the GKE_ZONE, GKE_CLUSTER, IMAGE, REGISTRY_HOSTNAME and DEPLOYMENT_NAME environment variables (below).
```
imjohnbo, line 9:
```suggestion
# 3. Change the values for the GKE_ZONE, GKE_CLUSTER, IMAGE, and DEPLOYMENT_NAME environment variables (below).
```

ci/google.yml
imjohnbo: It would be best to reference a major version like `v1` [[docs](https://docs.github.com/en/actions/creating-actions/about-actions#using-release-management-for-actions)]

ktoor-google: I don't think we have a v1 tagged release. Would it be recommended to pin it to 0.1.3 or keep it master?
```suggestion
    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@v0.1.3
```

imjohnbo: Yeah I think `0.1.3` is fine for now as a replacement for `master` but it will quickly become out of date as new features (and actions) are added, which is why we recommend pinning to a major version tag like `v1` (`v` is an optional recommended prefix) and moving that tag to the Gif ref of the current release.

```suggestion
    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.1.3
```

ktoor-google:
All set @imjohnbo @bharathkkb

PR #622, opened 2020-08-10
Maven no transfer progress

sullis:
This repository contains configuration for what users see when they click on the `Actions` tab.

It is not:

---

Thank you for sending in this pull request. Please make sure you take a look at the [contributing file](https://github.com/actions/starter-workflows/blob/main/CONTRIBUTING.md). Here's a few things for you to consider in this pull request.

- [ ] A good description of the workflow at the bottom of this page.
- [ ] Some links to the language or tool will be nice (unless its really obvious)

Some general notes:

    ```
    # documentation.
    ```


Marcono1234:
If I understand [MNG-6605](https://issues.apache.org/jira/browse/MNG-6605), which added this option, correctly it is only needed when running in interactive mode, but not when running in batch mode (`-B`). So it appears this might be redundant.

sullis:
> If I understand [MNG-6605](https://issues.apache.org/jira/browse/MNG-6605), which added this option, correctly it is only needed when running in interactive mode, but not when running in batch mode (`-B`). So it appears this might be redundant.

Here is an example build:
```
mvn -B clean package
```

Maven's output includes network transfer activity:

Screen Shot 2020-11-25 at 2 22 29 PM

joshmgross (commented):
👋  Hey @sullis, could you help me understand why we should add this to the starter workflows?

If there are download issues during the maven step, wouldn't users want to see the transfer progress?

Marcono1234:
> If there are download issues during the maven step, wouldn't users want to see the transfer progress?

Not the original reporter; but it appears that Maven still informs you about download errors with `--no-transfer-progress` (if the download failed for all configured repositories), and even informs you about warnings when the update policy requires checking for new versions (e.g. for snapshot dependencies), but  fetching artifact metadata failed.

So the point of this pull request is likely to reduce the amount of clutter when a project has a lot of dependencies which need to be downloaded.
Though note that when using the GitHub cache action to [cache Maven dependencies](https://github.com/actions/cache/blob/main/examples.md#java---maven), this might only be a problem for the initial build; afterwards the majority (or all) dependencies have already been downloaded.

PR #625, opened 2020-08-13
npm-publish / published release instead of created

cvergne:
Hi,

A very little change on the trigger event from `created` release to `published` to avoid package published when a release draft is created

Because currently, it will publish each time a draft is created, which is not what we want usually :)

Note that I used `published` instead of `released` to include `pre-release`

### DIFF
--- a/ci/npm-publish.yml
+++ b/ci/npm-publish.yml
@@ -5,7 +5,7 @@ name: Node.js Package
 
 on:
   release:
-    types: [created]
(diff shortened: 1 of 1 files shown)


kadirselcuk (approved):

cvergne:
Hi,
As it has been automatically closed due to no activity on PR, I open it again as I think it's still relevant. Tell me if it's not :)

Thanks !

PR #618, opened 2020-08-08
Update workflows to use `pull_request_target` event

ylemkimon:
GitHub [has introduced](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) a new event type: `pull_request_target`, which allows to run workflows from base branch and pass a token with write permission.

> In order to solve this, we’ve added a new `pull_request_target` event, which behaves in an almost identical way to the `pull_request` event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request. This event can be used in combination with the private repository settings as well.

PR #626, opened 2020-08-14
Add warning: labeler action nonfunctional on pulls from forks

HonkingGoose:
**EDIT:** I think it would be better to merge #618, as that fixes the problem instead of adding a note like my pull.
I'll leave it up to the maintainers to decide what they like best. :smile: 

---

**Changes**

- Add warning about the _labeler_ action not processing pull requests coming from forks.

**Context**

I needed to look in the docs for the _labeler_ action itself to find out it will not label pull requests coming from forks.
This information should be presented prior to the user committing it to their repository, and getting confused about why forked pull requests are not labeled.

Relevant snippet from [Labeler action readme](https://github.com/actions/labeler/blob/main/README.md):
> Note that only pull requests being opened from the same repository can be labeled. This action will not currently work for pull requests from forks -- like is common in open source projects -- because the token for forked pull request workflows does not have write permissions.


ylemkimon:
Note it's now possible to use the Labeler action with PRs from forks, see #618.

HonkingGoose:
Closing this in favor of #618.

PR #632, opened 2020-08-21
updates to starter workflows

andymckay:
This is an update to the policy around new community starter workflows as mentioned in https://github.com/github/c2c-actions/issues/1216


alauq (commented):
### CONTRIBUTING.md

PR #637, opened 2020-08-30
obtain files with path, ignore build

### LINKED ISSUE #636: pylint action is broken
In https://github.com/actions/starter-workflows/blob/main/ci/pylint.yml, the last line:

```shell
pylint `ls -R|grep .py$|xargs`
```

does not work as intended. `ls -R | grep .py$` lists all files with the `.py` ending. Only the *files* not the path to them. So it will find `.py` files in a subdirectory, and pass them to pylint which is then unable to find them on the top level. Causing errors like this:

```bash
vscode@5aa8d281d9b3:/workspaces/lazyConfig$ ls -R|grep .py$
setup.py
__init__.py
config.py
lazyData.py
__init__.py
config.py
lazyData.py
__init__.py
smoke_test.py
vscode@5aa8d281d9b3:/workspaces/lazyConfig$ pylint config.py
************* Module config.py
config.py:1:0: F0001: No module named config.py (fatal)
vscode@5aa8d281d9b3:/workspaces/lazyConfig$ 
```

I'll see if I can fix this and open a pull request if I can find a good fix.

FelixBenning:
This is an attempt to fix #636 

I am unsure what the difference between `$()` and ` `` ` is. I am more used to `$()`, but I actually do not know the difference. So this is not a thoughtful change.

I also noticed that it would list files in `./build` locally, which are of course duplicates. So I removed them with the `-not -path`. In general those files should never be on github as they should be in `.gitignore`. So maybe this is unnecessary.


FelixBenning:
It appears that `pylint` uses non zero exit codes liberally. `pylint-exit` filters these codes and returns a zero exit code on non critical things.

andymckay:
Both versions of pylint worked for me in very simple testing. @fjy202 wrote the original starter workflow, so maybe they have more insight here.

FelixBenning:
have you created a module?
this is the old command `ls -R|grep .py$|xargs`

`ls -R` only prints the filename not the (relative) path. So if you have modules in folders, then the filename itself is not a valid argument for `pylint` so it fails.

E.g. `folder/example.py` would result in this: `pylint example.py`

FelixBenning:
@andymckay is there interest in this? otherwise I will not resolve merge conflicts.

ci/pylint.yml
```diff
@@ -17,6 +17,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install pylint
-    - name: Test with pytest
+        pip install pylint-exit
+    - name: Run Pylint
       run: |
-        pylint `ls -R|grep .py$|xargs`
+        pylint $(find . -type f -name "*.py" -not -path './build*' | xargs) || pylint-exit $?
```
gpoo, line 23: A simpler solution is to use to call `pylint` directly from find:

```suggestion
        find -regextype egrep -regex '(.*.py)$' -exec pylint {} \;
```

I use egrep type with that format because it can be extended, for example, to

```sh
find -regextype egrep -regex '(.*.py|.*.py.in)$' -exec pylint {} \;
```

I do not see the purpose of excluding the `build` directory. In which circumstances would you add the `build` to the repository?

gpoo:
> This is an attempt to fix #636
> 
> I am unsure what the difference between `$()` and `` `` `` is. I am more used to `$()`, but I actually do not know the difference. So this is not a thoughtful change.

`$()` is a bash-ism (also available in zsh, and perhaps adopted by other shells). Backticks is more universal across different shells.

FelixBenning:
@gpoo I see you also removed the use of `pylint-exit`. I personally don't like a linter causing a hard fail for anything it finds, but at the same time it is easy to append that while it might be less noticable if it just passes without a notification so there are pros and cons. Not something I feel strongly about.

gpoo:
> @gpoo I see you also removed the use of `pylint-exit`. I personally don't like a linter causing a hard fail for anything it finds, but at the same time it is easy to append that while it might be less noticable if it just passes without a notification so there are pros and cons. Not something I feel strongly about.

It was a suggestion. After second thoughts, using find maybe wrong as the output will be whether `find` was successful or not, and that maybe misleading, as one might expect the result of `pylint`.

As far as I understand, pylint-exit is a separate utility, so it would need to be installed.

Shaynefrost (approved):
713e3be504c947026457a5b8690a1137557e8ba6

FelixBenning:
@gpoo 
> After second thoughts, using find maybe wrong as the output will be whether find was successful or not, and that maybe misleading, as one might expect the result of pylint.

I tested it and it does in fact eat all exit codes. So the exit code is always 0. I made another version - here it is quite obvious what you need to change to get a zero exit code if the code fails. It also saves an artifact.

It would also be nice if you could somehow read out the previous artifact and fail under the previous score

080708km:
> This is an attempt to fix #636
> 
> I am unsure what the difference between `$()` and `` `` `` is. I am more used to `$()`, but I actually do not know the difference. So this is not a thoughtful change.
> 
> I also noticed that it would list files in `./build` locally, which are of course duplicates. So I removed them with the `-not -path`. In general those files should never be on github as they should be in `.gitignore`. So maybe this is unnecessary.
Okok

Toppole69 (commented):
H

jDavidnet:
This would be helpful to merge into `main`.  the current script doesn't work on any python scripts in a subdirectory.  It both can't lint these files, and produces an error code of `F0001: No module named`

Without this fix, it's nearly worthless.

ServrBig4uNet (approved):

ci/pylint.yml
```diff
@@ -19,4 +19,10 @@ jobs:
         pip install pylint
     - name: Analysing the code with pylint
       run: |
-        pylint `ls -R|grep .py$|xargs`
```
Saiful79NZ, line 22: It easy if you could make it larger scale or just make it auto

Saiful79NZ (approved):

baraBlz (approved):
Thank you. @Saiful79NZ

Alenapapa (approved):

GetmyMoney (commented):
Fuck all of you I got you now

baraBlz (approved):

ci/pylint.yml
HUMM1984: Thanks

HUMM1984 (approved):

baraBlz (approved):

mehran-sayyah:
#637

HUMM1984 (commented):
Looks good

PR #639, opened 2020-09-01
Update Publish Docker template to publish to GitHub Container Registry

kaylangan:
GitHub Container Registry (GHCR) is intended to supersede the Docker service in GitHub Packages. I've updated the Publish Docker template to use GHCR which was released as a public beta on 9/1/20:

- GHCR doesn't support `GITHUB_TOKEN` so a PAT will need to be created and saved as `CR_PAT` instead.
- The URL is ghcr.io instead of docker.pkg.github.com.
- The repo path part is no longer required. You publish to ghcr.io/[org]/[image-name].

### DIFF
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -50,12 +50,13 @@ jobs:
       - name: Build image
         run: docker build . --file Dockerfile --tag $IMAGE_NAME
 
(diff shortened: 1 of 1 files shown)


Zenexer:
Why was this reverted?  Is there an alternative that should be used?

PR #646, opened 2020-09-19
Change parameter to pylint

rupakbajgain:
Previous operation does not actually do work recursively and this works for windows too.
can also use pylint $(git ls-files '*.py')

### DIFF
--- a/ci/pylint.yml
+++ b/ci/pylint.yml
@@ -19,4 +19,4 @@ jobs:
         pip install pylint
     - name: Analysing the code with pylint
       run: |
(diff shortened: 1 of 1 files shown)


rupakbajgain:
`ls -R|grep .py$|xargs` gives file name only no path

Kraisornbiinthi (commented):
Thank you

rupakbajgain:
@Kraisornbiinthi  I have closed this PR bcz `pylint **/*.py` also don't check every file but first search for module with __init__.py only.

PR #647, opened 2020-09-20
Update pylint.yml

rupakbajgain:
`ls -R|grep .py$|xargs` gives file name only no path

### DIFF
--- a/ci/pylint.yml
+++ b/ci/pylint.yml
@@ -19,4 +19,4 @@ jobs:
         pip install pylint
     - name: Analysing the code with pylint
       run: |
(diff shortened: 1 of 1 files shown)


ci/pylint.yml
```diff
@@ -19,4 +19,4 @@ jobs:
         pip install pylint
     - name: Analysing the code with pylint
       run: |
-        pylint `ls -R|grep .py$|xargs`
```
Kraisornbiinthi, line 22: I

Kraisornbiinthi (approved):

PR #681, opened 2020-10-22
add template workflows for code scanning

nickfyson:
This PR adds template workflows for Code Scanning, and contains those workflows that are already presented in the setup page. They're qualitatively different to those workflows already present in the repo, and hence we add a new folder `code-scanning`.

- - -

Points from the template to address...

    ```
    # documentation.
    ```


nickfyson:
Closing in favour of separate PRs, one per partner workflow.

PR #685, opened 2020-10-27
Get rid of quoting that breaks by default.

Sheeri:
It looks like someone tried to use '' to make a ' happen for the possessive tense of users. But that messes up the quoting:
issue-message: 'Message that will be displayed on users'' first issue'
pr-message: 'Message that will be displayed on users'' first pr'

It was a very bad impression to have a simple script designed to welcome folks be broken by default.

It should be:
issue-message: 'Message that will be displayed on the first issue for that user'
pr-message: 'Message that will be displayed on the first pr for that user'

(this gets rid of the spurious quotes, but also doesn't introduce any grammatical errors)

-------

Some general notes:

    ```
    # documentation.
    ```


automation/greetings.yml
```diff
@@ -9,5 +9,5 @@ jobs:
     - uses: actions/first-interaction@v1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        issue-message: 'Message that will be displayed on users'' first issue'
-        pr-message: 'Message that will be displayed on users'' first pr'
+        issue-message: 'Message that will be displayed on users'
+        pr-message: 'Message that will be displayed on users'
```
andymckay, line 13:
```suggestion
        issue-message: 'Message that will be displayed on users first issue'
        pr-message: 'Message that will be displayed on users first pull request'
```

andymckay (approved):
Thanks for spotting this @Sheeri, I agree. Can we add in my suggestion, without it the sentence just trails off and doesn't make sense.

automation/greetings.yml
Sheeri: Hi @andymckay - Indeed, we don't want the sentence to trail off. However, now it's gramatically incorrect. How about:

issue-message: 'Message that will be displayed on the first issue for that user'
pr-message: 'Message that will be displayed on the first pr for that user'

andymckay: Sure that works for me 👍🏾

andymckay:
Added suggestions and merged, so we can remove this breakage for new users. Hope that's ok @Sheeri.

Sheeri:
absolutely, thanks!

PR #690, opened 2020-10-30
Update dart.yml

devoncarew:
This PR updates the github actions starter workflow for Dart.

Note that I am a member of the Dart team and am more than happy to answer any questions about these changes.

This PR:
- discusses the other release channels we have available (dev and beta, in addition to the existing `google/dart:latest` stable channel docker image referenced below), as well as includes a link where users can find the list of docker images
- adds a code analysis step
- adds a format verification step
- the format step is commented out as that's more likely to depend on the workflow of specific teams (but useful to know the command line to run)
- updates the test step to use our newer `dart test` tooling (instead of the existing `pub run test` command)
- for the test step, mention the slightly different command that a Flutter project will need to use

Thanks!

---

**Please note that we are not accepting new starter workflows at this time. Updates to existing starter workflows are fine.**

---

Some general notes:

    ```
    # documentation.
    ```

### DIFF
--- a/ci/dart.yml
+++ b/ci/dart.yml
@@ -20,22 +20,22 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: dart --version
+      - name: Print Dart SDK version
         run: dart --version
 
-      - name: dart pub get
+      - name: Install dependencies
         run: dart pub get
 
       # Uncomment this step to verify the use of 'dart format' on each commit.
-      # - name: dart format
+      # - name: Verify formatting
       #   run: dart format --output=none --set-exit-if-changed .
 
       # Consider passing '--fatal-infos' for slightly stricter analysis.
-      - name: dart analyze
+      - name: Analyze project source
         run: dart analyze
 
       # Your project will need to have tests in test/ and a dependency on
       # package:test for this step to succeed. Note that Flutter projects will
       # want to change this to 'flutter test'.
-      - name: dart test
+      - name: Run tests
         run: dart test
--- a/ci/dart.yml
+++ b/ci/dart.yml
@@ -1,4 +1,4 @@
-name: Dart CI
+name: Dart
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


ci/dart.yml
```diff
@@ -8,15 +8,30 @@ on:
...
     steps:
-    - uses: actions/checkout@v2
-    - name: Install dependencies
-      run: pub get
-    - name: Run tests
-      run: pub run test
+      - uses: actions/checkout@v2
+
+      - name: Install package dependencies
+        run: pub get
```
mit-mit, line 24: `dart pub get`

ci/dart.yml
```diff
@@ -8,15 +8,30 @@ on:
...
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Install dependencies
-      run: pub get
-    - name: Run tests
-      run: pub run test
+      - uses: actions/checkout@v2
+
+      - name: Install package dependencies
```
mit-mit, line 23: I've found it useful to add this step at the first step:

```
      - name: Report version
        run: dart --version
```

ci/dart.yml
```diff
@@ -8,15 +8,30 @@ on:
...
+        run: pub get
+
+      # Comment this step in to verify the use of 'dart format' on each commit.
```
mit-mit, line 26: "Comment ... in" is a bit confusing; consider "# Uncomment ..."

mit-mit (changes requested):

ci/dart.yml
devoncarew: Changed

ci/dart.yml
devoncarew: updated

ci/dart.yml
devoncarew: added

devoncarew (commented):
Thanks for the review! Updated.

mit-mit:
Great. Why the change to make all the step names are the same as the commands?

devoncarew:
> Great. Why the change to make all the step names are the same as the commands?

When I've done this on CIs (https://github.com/dart-lang/usage/runs/1334831476?check_suite_focus=true, ...) I think it reads a bit better.

mit-mit:
Hmm, it seems a bit inconsistent with "Setup job" and "Initialize containers"?

devoncarew:
> Hmm, it seems a bit inconsistent with "Setup job" and "Initialize containers"?

Yes, likely, but I think it helps clarify to Dart users what the various tasks are doing (there will be less guessing about what's in the 'pub get' or 'dart analyze' sections than there would be for 'Install package dependencies' or 'Analyze project' sections).

andymckay (commented):
This looks fine to me, but I agree with name changes. Looking across some of the other starter workflows in this repository, this is not a pattern that's common. Let's keep it the same as the rest with a generic description of what the step is about.

devoncarew:
> Let's keep it the same as the rest with a generic description of what the step is about.

Sure, happy to revert the name changes. I'll update the names and ping this PR again.

devoncarew (commented):
PTAL - I revered the step names, and used sentence case for the new step names (`Verify formatting`, ...).

andymckay (approved):

andymckay:
Thanks for the udpates @mit-mit 🙇🏾

PR #694, opened 2020-11-02
Use GitHub-Recommened Endpoint

developStorm:
Although I haven't tested the differences between those two endpoints in detail yet, this recommended one does look like the only working one now on my end. 

image


developStorm:
Was fixed in f75012de65c4d2e22b548521929b053e5c929737, close this then.

PR #697, opened 2020-11-04
remove misleading info

xiongnemo:
python-django.* has renamed to django.*, while the doc remains the same


andymckay (approved):
Thank you 🙇🏾

Issue #631, opened 2020-08-21
Pause on accepting on new community starter workflows


andymckay:
👋🏾 Hi there

As you can probably tell, we haven't been very good at merging in new starter workflows in to this repository. First off I'd like to apologise to everyone who has submitted a new starter workflow. Thank you for submitting them and our apologies for not accepting them sooner.

We'd love to have more starter workflows appear on the Actions tab, but at this time the page just doesn't work well with lots of new starter workflows. Having a team try to process and understand the starter workflows hasn't worked too well.  

Perhaps we need to make starter workflows more like the Marketplace, so a community can publish and update workflows without needing approval. That's something we are exploring for the future.

At this time I think its unfair to starter workflow authors to submit them and then have nothing happen with them for a long period of time. Rather than create unfair expectations we are temporarily going to **stop accepting new** starter workflows. We **will try to merge all the ones already submitted**, see below 👇🏾 for that list.

Sorry about that, but hope this is temporary and we'll figure something else out that's better for the future.

This are the starter workflows we are going to try and merge in:

- [x] https://github.com/actions/starter-workflows/pull/605 or https://github.com/actions/starter-workflows/pull/580 Nix - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/604 Deno - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/591 Phoenix - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/587 Ionic - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/576 or https://github.com/actions/starter-workflows/pull/489 Open shift - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/546 or https://github.com/actions/starter-workflows/pull/74 D Lang
- [x] https://github.com/actions/starter-workflows/pull/526 Symfony - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/493 Ada - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/460 iOS - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/329 Hakyll - 😢 
- [x] https://github.com/actions/starter-workflows/pull/321 ASP.NET core deployment to Azure - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/311 Cypress - changes requested.
- [x] https://github.com/actions/starter-workflows/pull/298 Pylint
[...]

PR #700, opened 2020-11-05
Basic GitHub workflow for a Perl distribution

atoomic:
Add GitHub workflow for Perl distributions

### DIFF
--- a/ci/perl.yml
+++ b/ci/perl.yml
@@ -1,0 +1,103 @@
+name: Perl testsuite
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


andymckay:
👋🏾 I think this falls under this issue here of not accepting new starter workflows #631. Please let me know if that's incorrect.

atoomic:
indeed, what would be the best place to monitor when that policy change?
thanks

andymckay:
👍🏾 This repository, I'll put an update on issue #631.

PR #703, opened 2020-11-10
Prepare python-package.yml for python 3.10

boxed:
Python 3.10 is coming soon, and this will cause problems with the code as currently written. The python versions are written as floats and not strings, which will mean that 3.10 == 3.1, which is going to be very surprising.


sohoskool (approved):
Approve changes

sohoskool (approved):
Approved

sohoskool (changes requested):
Request and Approve Changes

andymckay (approved):
Thank you 🙇🏾 In YAML `3.1` will be the same as `3.10` so that makes sense we should make them strings 👍🏾

PR #713, opened 2020-11-19
Add workflow_dispatch trigger to the default empty workflow

cschleiden:
This repository contains configuration for what users see when they click on the `Actions` tab and the setup page for Code Scanning.

It is not:

---

**Please note that at this time we are only accepting new starter workflows for Code Scanning. Updates to existing starter workflows are fine.**

---

- [x] Code Scanning workflows should run on `push` to `branches: [ $default-branch, $protected-branches ]` and `pull_request` to `branches: [ $default-branch ]`. We also recommend a `schedule` trigger of `cron: $cron-weekly`.

Some general notes:

- [x] This workflow must only use actions that are produced by the language or ecosystem that the workflow supports.  These actions must be [published to the GitHub Marketplace](https://github.com/marketplace?type=actions).  We recommend that these actions be referenced using the full 40 character hash of the action's commit instead of a tag.  Additionally, workflows must include the following comment at the top of the workflow file:
    ```
    # documentation.
    ```

### DIFF
--- a/ci/blank.yml
+++ b/ci/blank.yml
@@ -10,7 +10,7 @@ on:
   pull_request:
     branches: [ $default-branch ]
 
-  # This allows you to run this workflow manually from the Actions tab
+  # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
--- a/ci/blank.yml
+++ b/ci/blank.yml
@@ -2,13 +2,16 @@
 
 name: CI
 
-# Controls when the action will run. Triggers the workflow on push or pull request
-# events but only for the $default-branch branch
+# Controls when the action will run. 
 on:
+  # Triggers the workflow on push or pull request events but only for the $default-branch branch
   push:
     branches: [ $default-branch ]
   pull_request:
     branches: [ $default-branch ]
+
+  # This allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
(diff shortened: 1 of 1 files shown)


ci/blank.yml
```diff
@@ -2,14 +2,17 @@
 
 name: CI
 
-# Controls when the action will run. Triggers the workflow on push or pull request
-# events but only for the $default-branch branch
+# Controls when the action will run. 
 on:
+  # Triggers the workflow on push or pull request events but only for the $default-branch branch
```
mscoutermarsh, line 7:
```suggestion
  # Triggers the workflow on push or pull request events but only for the default branch
```

Maybe? Reads a little easier.

mscoutermarsh (approved):

ci/blank.yml
cschleiden: @mscoutermarsh `$default-branch` is automatically replaced with the name of the actual default branch when you use this workflow

ci/blank.yml
```diff
@@ -2,14 +2,17 @@
...
   push:
     branches: [ $default-branch ]
   pull_request:
     branches: [ $default-branch ]
 
+  # This allows you to run this workflow manually from the Actions tab
```
cschleiden, line 13: Maybe just? 
```suggestion
  # Allows you to run this workflow manually from the Actions tab
```

ci/blank.yml
mscoutermarsh: Yeah I know, I read it in my head as "default branch default branch"

andymckay (approved):
Looks good, let's do it 🚢 👍🏾

PR #717, opened 2020-11-23
Use consistent calls to pip

RobertRosca:
`pip` was called in two different ways which is a bit inconsistent, this change just makes it so that it is called in the same way both times.


andymckay (approved):
Thanks 🙇🏾 looks good.

PR #720, opened 2020-11-24
Build/Test all go packages

ostaebler:
Use the "..." pattern to build/test all go packages in a repository.


andymckay (approved):
Thanks 🙇🏾‍♂️ this looks good.

Issue #539, opened 2020-06-03
Improve "maven-publish.yml"


Marcono1234:
Hello,
[`maven-publish.yml`](https://github.com/actions/starter-workflows/blob/master/ci/maven-publish.yml) could probably be improved:
1. Improve command line parameters, see also #538
2. Link to https://github.com/actions/setup-java#apache-maven-with-a-settings-path should be placed directly above `settings-path:` because it is irrelevant for everything else
3. Add link to https://github.com/actions/setup-java#publishing-using-apache-maven in header comment (or to [GitHub Help: Publishing Java packages with Maven](https://help.github.com/en/actions/language-and-framework-guides/publishing-java-packages-with-maven))
4. Remove redundant step `name: Build with Maven`, `deploy` already includes `package` so it essentially runs `package` (and all phases before it) twice
Edit: `deploy` would skip compiling (since `package` already did this), but would run tests again.
5. Add `--batch-mode` to `deploy` to prevent output which is intended for interactive console
6. Write out `-s` as `--settings` to make it easier to understand for users who are not familiar with all command line options
7. Change name "Publish to GitHub Packages Apache Maven" to "Publish to GitHub Packages with Maven" to be consistent with other message
8. It might be possible to run `deploy` with `-Dmaven.install.skip=true` to skip installing the package to the local repository since that could cause issues when a cache step is used and a large number of the packages of the project accumulate (only happens if the version of the package changes frequently). Though would be good if someone with more Maven deployment experience could confirm this.

Let me know if you want me to write a pull request.

See also https://github.com/actions/setup-java/issues/75

andymckay:
I'm not a maven user, but a pull request for these changes would be most welcome.

PR #721, opened 2020-11-24
Create stale.yml

andymckay:
We've got lots of stale issues and PRs on this repository. This will hopefully help clean them up.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "21 4 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has become stale and will be closed automatically within a period of time. Sorry about that.'
+        stale-pr-message: 'This pull request has become stale and will be closed automatically within a period of time. Sorry about that.'
+        stale-issue-label: 'no-issue-activity'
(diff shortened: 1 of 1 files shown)


cschleiden (approved):

Marcono1234:
Could you please reconsider this? In my opinion the default configuration of the `stale` action is in its current form contribution-hostile (see also https://github.com/actions/stale/issues/56).

For example for #538 and #539 the bot wants to close the issue / pull request without any interaction from one of the contributors / maintainers. If you need more information, want any changes or don't like the changes please let me know. But having the bot just close them is really not nice.

Maybe a better solution would be to use the `only-labels` option of the `stale` action and then label any issue / pull request where you are awaiting a user response.

andymckay:
Thank you for pointing out that PR and issue, I've commented on or merged those.

A reduction in the number of issues and PRs would be an overall good for this repository, especially given the amount of spam and bots that write into this repository. Stale sends a ping to the watchers and let's them review the issue or PR when it occurs. At this point, I don't see a need to change that policy.

Marcono1234:
Thanks for the insight, I was not aware of the spam. Though I am worried that the situation becomes as bad as for https://github.com/actions/stale where the bot closed a lot of good issues. And indeed it appears the situation here is similar because most issues labelled `no-issue-activity` and some pull requests labelled `no-pr-activity` appear to have been inactive due to missing interaction by you, the maintainers 😕

Would an alternative be to have a label for good issues and pull requests, e.g. `triaged` (or alternatively use an issue template which adds a `non-triaged` label) and then one for issues and pull requests where you are expecting a response?
Then the bot could close all issues and pull requests without `triaged` label, or with `awaiting-response` label. This would still require that you triage them in time, but that would only have to be checking for non-spam (though maybe that would defeat the addition of the bot to tackle spam).

PR #727, opened 2020-12-01
Update ruby.yml

andymckay:
Noticed that this version of setup-ruby causes problems with the `set-env` deprecation. This just bumps it up to the latest commit on https://github.com/ruby/setup-ruby/


andymckay:
cc @eregon

mscoutermarsh (approved):

andymckay:
Just going to merge this, @eregon if there's any concerns, we can use a different hash.

eregon:
That looks good, thanks.

PR #731, opened 2020-12-05
Add conda to system path

davidar:
This cleans up the actions a little, but more importantly sets up the environment more similarly to `conda activate`.

---

- [X] The workflow filename of CI workflows should be the name of the language or platform, in lower case.  Special characters should be removed or replaced with words as appropriate (for example, "dotnet" instead of ".NET").

- [X] Includes a matching `ci/properties/*.properties.json` file.
- [X] Use sentence case for the names of workflows and steps, for example "Run tests".
- [X] The name of CI workflows should only be the name of the language or platform: for example "Go" (not "Go CI" or "Go Build")
- [X] Include comments in the workflow for any parts that are not obvious or could use clarification.
- [ ] Code Scanning workflows should run on `push` to `branches: [ $default-branch, $protected-branches ]` and `pull_request` to `branches: [ $default-branch ]`. We also recommend a `schedule` trigger of `cron: $cron-weekly`.

Some general notes:

- [X] This workflow must only use actions that are produced by GitHub, [in the `actions` organization](https://github.com/actions), **or**
- [ ] This workflow must only use actions that are produced by the language or ecosystem that the workflow supports.  These actions must be [published to the GitHub Marketplace](https://github.com/marketplace?type=actions).  We recommend that these actions be referenced using the full 40 character hash of the action's commit instead of a tag.  Additionally, workflows must include the following comment at the top of the workflow file:
    ```
    # documentation.
    ```

### DIFF
--- a/ci/python-package-conda.yml
+++ b/ci/python-package-conda.yml
@@ -14,18 +14,21 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
-    - name: Install dependencies
+    - name: Add conda to system path
       run: |
         # $CONDA is an environment variable pointing to the root of the miniconda directory
-        $CONDA/bin/conda env update --file environment.yml --name base
+        echo $CONDA/bin >> $GITHUB_PATH
+    - name: Install dependencies
+      run: |
+        conda env update --file environment.yml --name base
     - name: Lint with flake8
       run: |
-        $CONDA/bin/conda install flake8
+        conda install flake8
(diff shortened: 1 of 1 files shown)


mjguidry68 (approved):

mjguidry68 (approved):

andymckay (approved):
Looks good, thanks 👍🏾

O9864 (approved):

PR #738, opened 2020-12-07
Improve maven-publish.yml

### LINKED ISSUE #539: Improve "maven-publish.yml"

Marcono1234:
Fixes #539

- Remove redundant build step (`deploy` covers that anyways)
- Use long version of Maven command line parameter names
- Add `maven.install.skip=true` to skip installing artifact to local repository
Since the purpose of this workflow is to deploy the artifact, there is usually no need to install it to the local Maven repository. Though maybe there are use cases where this is desired; please let me know whether I should omit that.

In my private testing repository this modified workflow seems to work fine, but it would be good if someone more familiar with Maven could have a look at as well.

### DIFF
--- a/ci/maven-publish.yml
+++ b/ci/maven-publish.yml
@@ -1,14 +1,14 @@
 # This workflow will build a package using Maven and then publish it to GitHub packages when a release is created
-# For more information see: https://github.com/actions/setup-java#apache-maven-with-a-settings-path
+# For more information see: https://docs.github.com/en/free-pro-team@latest/actions/guides/publishing-java-packages-with-maven#publishing-packages-to-github-packages
 
-name: Maven Package
+name: Maven Publish to GitHub
 
 on:
   release:
     types: [created]
 
 jobs:
-  build:
+  deploy:
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


newblaque (approved):

kukulkankanpull (changes requested):
@Marcono1234  review required

kukulkankanpull (approved):
`#539`

kukulkankanpull (approved):

kukulkankanpull:
Marcono1234:marcono1234/maven-publish

kukulkankanpull (approved):

andymckay (approved):
Looks good to me @Marcono1234, I don't have maven experience enough to test it out though.

johnperez416 (approved):

johnperez416 (approved):

kabat87 (approved):

breeze87j (approved):

PR #723, opened 2020-11-25
dotnet-Workflows: Use latest channel version

litetex:
Fixes #619 

Originally fixed by #620 
This version was updated to .NET 5

PR #740, opened 2020-12-10
Dotnet updates

rainersigwald:
I set up a new repo from the template today and noticed a few issues with the workflow:

* Stale branding (".NET Core" is just ".NET" going into the future).
* Confusing naming of restore step
* Restore/Build/Test were split up but not consistently using the results of the prior step.

Some more details in individual commit messages.

I didn't update to the latest released SDK since that's done in #723 (blocked on agent updates).

Feedback from @kathleendollard and @timheuer on this would probably be useful.


ci/dotnet.yml
```diff
@@ -13,13 +13,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Setup .NET Core
+    - name: Setup .NET
```
timheuer, line 16: If we were to suggest this change, then we should also bump dotnet-version as well as 3.1.301 is still 'NET Core' branded  and just for consistency if we are nudging it along.

rainersigwald: That's #723. I'd be fine waiting on this change until that merges.

kukulkankanpull (approved):

andymckay:
Works for me, thanks @rainersigwald

andymckay (approved):

PR #762, opened 2020-12-28
fix context references in cmake template

### LINKED ISSUE #761: Error in CMake starter template
Using the CMake starter template results in a `Error: No such file or directory` error when running the `CMake Configure step`

This is because the `${{ runner.workspace }}` and `${GITHUB_WORKSPACE}` variables are not the same, eg in https://github.com/planetmarshall/libftdi/runs/1618983723

`${{ runner.workspace }}` is `/home/runner/work/libftdi` and `${GITHUB_WORKSPACE}` is `/home/runner/work/libftdi/libftdi`

planetmarshall:
Fixes #761


joshmgross (approved):
Thanks for this!

PR #764, opened 2021-01-02
haskell: ghc 8.10 and cabal 3.2

chenrui333:
keep using the latest haskell offerings when bootstrapping the action scripts.


andymckay (approved):

andymckay:
Thanks @chenrui333

PR #768, opened 2021-01-04
go: update starter

chenrui333:
update starter workflow

---

go 1.16 would [use go modules build by default](https://tip.golang.org/doc/go1.16#modules), so it is good for new projects to use without dep (also `dep` project was [archived last year](https://github.com/golang/dep/commit/f13583b555deaa6742f141a9c1185af947720d60)).


chenrui333:
I actually made several action PRs into the new golang projects recently as well, so I feel it would be good to change in here.

joshmgross (approved):

chenrui333:
@joshmgross merging this as well? :)

chenrui333:
Thanks @joshmgross!

PR #766, opened 2021-01-02
ruby: add ruby 2.7 and 3.0 into bootstrap script

chenrui333:
ruby: add ruby 2.7 and 3.0 into bootstrap script

---

relates to https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

PR #769, opened 2021-01-04
Recommend the bundler-cache option

eregon:
* It is both more convenient and significantly faster.
* It's also recommended in the action's README: https://github.com/ruby/setup-ruby/#usage


chenrui333:
relates to #766 (which has this change)

eregon:
@chenrui333 Thanks, I didn't see that PR. I'm OK to merge either, but if it's #766 let's squash it to avoid back-and-forth history and extra merges.

joshmgross:
Thanks! I went ahead and merged https://github.com/actions/starter-workflows/pull/766 which added this option.

PR #773, opened 2021-01-07
Fix grammar error

arnu515:
In the comment on line `1`, it says "This workflows", but I think that "This workflow" makes more sense.

This is just a simple grammar error and makes no changes to actual code.


joshmgross (approved):
Thanks!

afard:
Hi,
I added this workflow (python-publish.yml) to my repository (https://github.com/vertica/VerticaPy/blob/master/.github/workflows/python-publish.yml) and then created a release (0.4.1), but the workflow was not triggered. Can you please advise me?

PR #823, opened 2021-02-25
Use ruby/setup-ruby@v1 in Ruby starter workflow

eregon:
I am the main author of ruby/setup-ruby.
I think it's time to simply use `@v1` in this starter workflow.

Even the official documentation about this action, by GitHub, uses `ruby/setup-ruby@v1`:
https://docs.github.com/en/actions/guides/building-and-testing-ruby

Probably the rule below merits updating:

    ```
    # documentation.
    ```

Does that sound acceptable?

If people actually want maximum security and stability (at the expense of everything else), they should probably pin *everything* down to a full sha (also actions developed by GitHub, they depend on third party packages which might change and have new security issues), but that seems very extreme and inconvenient, so I guess extremely few users would do that, and if they do, they would be well aware to do it for all actions, not just "3rd-party" ones.

Previous discussions about this and how weird and problematic it is to recommend a full sha to people trying GitHub Actions: #448, #709 and probably more.

Starter workflows should make it easy to get started.
If they're more cumbersome to use than the snippets in the documentation, there seems to be a problem.

### DIFF
--- a/ci/ruby.yml
+++ b/ci/ruby.yml
@@ -24,10 +24,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up Ruby
-    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
-    # change this to (see https://github.com/ruby/setup-ruby#versioning):
-    # uses: ruby/setup-ruby@v1
-      uses: ruby/setup-ruby@473e4d8fe5dd94ee328fdfca9f8c9c7afc9dae5e
+      uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby-version }}
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically


Gjoker13 (commented):
[]()

andymckay:
I'm sorry @eregon I totally get the reasoning for this you aren't the first person to ask. It is a pain for everyone involved, but at this time we can't relax this restriction. We do have to do something to improve the story here and we know about this internally within the Actions team.

I keep hoping at some point we can relax this, but for this moment the guidance that I'm following from security is to keep pinning to 40 char shas. I will go and check with the documentation team on that one, thanks for the pointer.

Sorry once again, we'll get there.

itsmeee1:
I'm just tring to get info from 9312141228 and I don know I I did. Right or not

devkadirselcuk (approved):

itsmeee1 (approved):

ci/ruby.yml
```diff
@@ -24,10 +24,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up Ruby
-    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
-    # change this to (see https://github.com/ruby/setup-ruby#versioning):
-    # uses: ruby/setup-ruby@v1
-      uses: ruby/setup-ruby@473e4d8fe5dd94ee328fdfca9f8c9c7afc9dae5e
+      uses: ruby/setup-ruby@v1
       with:
```
itsmeee1, line 31: 9312141228

itsmeee1 (approved):

itsmeee1 (approved):

ci/ruby.yml
itsmeee1: I would do that and pay for it if you could give me some how to get number 9312141228File attached and everything without them knowing I will pay for that like still￼ For two days now and I still have not satisfy my phone nothing just slowing  my phone￼ don and no info

PR #824, opened 2021-02-26
Make the validation strict by default

Seldaek:
Composer maintainer here. 

I would recommend running validate in strict mode by default, as it makes it exit with a non-zero code when there is a warning. By default warnings are not treated as fatal and as such the process still returns 0, but in CI that often means stuff gets missed.


AverageComet250 (approved):

joshmgross (approved):
Thanks!

PR #851, opened 2021-03-18
Remove the deprecated "--no-suggest" option from the Composer install step

yakimun:
Hello!

`php.yml` CI workflow starts Composer with the deprecated `--no-suggest` option.

It was fixed in #490 and hasn't been reverted as far as I can see, but the current version of the workflow contains this deprecated option for some strange reason.

Let's try to remove it again!


AverageComet250 (approved):

andymckay (approved):
Thanks @yakimun.

yakimun:
Thanks for your review @andymckay! Do we need anything else to merge this?

andymckay:
Oops forgot to hit that button, thanks for reminding me.

PR #885, opened 2021-04-27
Update the version of tfsec action

owenrumney:
- Updating the version of the action being used
- Using the exit code of tfsec in the exit of the action

## Pre-requisites

---

---

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```

### DIFF
--- a/code-scanning/tfsec.yml
+++ b/code-scanning/tfsec.yml
@@ -26,7 +26,6 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Run tfsec
-        id: run_tfsec
         uses: tfsec/tfsec-sarif-action@bcba1239223e5b6e20680dfcd72e53fbce5aab90
         with:
           sarif_file: tfsec.sarif         
@@ -36,6 +35,3 @@ jobs:
         with:
           # Path to SARIF file relative to the root of the repository
           sarif_file: tfsec.sarif  
-
-      - name: Exit with tfsec code
-        run: exit ${{ steps.run_tfsec.outputs.tfsec-return-code}}
--- a/code-scanning/tfsec.yml
+++ b/code-scanning/tfsec.yml
@@ -4,7 +4,6 @@
 # documentation.
 
 name: tfsec
-
 on:
   push:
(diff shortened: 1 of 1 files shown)


ServrBig4uNet (approved):

ServrBig4uNet (approved):

ServrBig4uNet (approved):

ServrBig4uNet (approved):

nickfyson:
Hi @owenrumney! Thanks for this update to the workflow.

Could you help me understand the exit code change? Under what conditions does `tfsec` exit with a non-zero code? 🙂

owenrumney:
Hey @nickfyson - when tfsec finds issues in the code it returns with a none zero. Previously we were swallowing it so as not to interrupt the upload of the sarif report but then it occurred to me that for the status check to block a PR then the exit code needs to be non zero. I noted that protected branches can be configured whether to all merge when the status checks are failing so presumed it should be deferred to that setting?

TL;DR - if any tfsec issues are found in the code scan it will exit != 0

Happy to adjust if I've misunderstood how it should be working.

nickfyson:
Thanks! I just wanted to double check what the intention was, as we're actually working on some very related changes. We're going to introduce a setting which allows repo admins to configure at what level of severity they want the `Code scanning results / tool` checkrun to fail (Error, Warning or Note). My concern in merging your suggested change for the interim is that users will end up with workflow behaviour that is incompatible with this new setting, and might result in unexpected/confusing outcomes.

I'm afraid we can't commit to a timeframe more precise than 'this quarter' (so by the end of June), but I think it would probably be better to wait, and let users rely on that new setting in due course?

owenrumney:
Okay, so if I've understood correctly the new change you're implementing will use the output that is uploaded to the code scanning results to determine what to do based on thresholds set by the user?

So the tfsec check should run - send its SARIF results - then complete as successful? I can remove that step from the PR which is fine, I'm just not seeing how the end user knows that their were issues if the status check doesn't show as failed.

owenrumney:
No probs @nickfyson - that makes sense. I've updated the PR to only be the slightly updated version

PR #889, opened 2021-04-30
chore: update `stale.yml`

MichaelDeBoey:
`repo-token` defaults to `github.token`

https://github.com/actions/stale/blob/043fbbdea3e060a5aa2949408ad8e4b52988f420/action.yml#L5-L8


andymckay (approved):
Good catch, thank you @MichaelDeBoey

PR #892, opened 2021-05-02
Do not use fail-fast when testing multiple Pythons

abitrolly:
Because the whole purpose of this worflow is to see on which Python the package fails (if it fails), and with `fail-fast` it leads to unnecessary re-runs.


andymckay (approved):
Makes sense @abitrolly

andymckay:
Thank you! 🙇🏾

PR #899, opened 2021-05-06
Add node v16 and remove v10 and v15

Mirobit:
Node v16 was released 2 weeks ago and v10 reached EOL a week ago. v15 will reach EOL in 3 weeks.
https://nodejs.org/en/about/releases/


anywherepilot (approved):

anywherepilot:
Thank you for noticing and taking action @Mirobit

PR #900, opened 2021-05-06
Update Python package build and upload example

jamescurtin:
## Pre-requisites

---

---

## Tasks

    - **Not Applicable**
    - **See summary below describing alternative release types**

**Some general notes:**

    - **Not Applicable**
    ```
    # documentation.
    ```

## Summary

The [Python Packaging Authority](https://www.pypa.io/en/latest/) (PyPA)--the working group responsible for `pip` and `PyPI`, among other packaging tools--maintains an action for uploading Python packages to PyPI.

This PR makes the following changes to the `python-publish` example:

1. Use the PyPA-maintained PyPI upload action instead of directly interacting with `twine`.
1. Uses [`build`](https://pypa-build.readthedocs.io/en/latest/) to generate the sdist & wheel that will be uploaded to PyPI instead of setuptools, as the former is [PEP-517](https://www.python.org/dev/peps/pep-0517/) compliant.
1. Uses a PyPI token instead of a username and password. Tokens may be scoped down to individual projects on PyPI, whereas the username and password can be used to upload artifacts for any package maintained by the user. As a result, using tokens is more secure.
1. Executes the action only on `release` events of the `released` or `prereleased` types. When a user makes a standard release, this will upload the package to PyPI. When the user makes a pre-release, this will upload the package to the development version of PyPI. This allows for testing the release process without permanently uploading new versions of a package to production PyPI. 

This [example repository](https://github.com/jamescurtin/test-package/actions) was used to demonstrate how this action will be invoked.

### DIFF
--- a/ci/python-publish.yml
+++ b/ci/python-publish.yml
@@ -1,11 +1,16 @@
 # This workflow will upload a Python Package using Twine when a release is created
 # For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
 
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
 name: Upload Python Package
 
 on:
   release:
-    types: [created]
+    types: [released, prereleased]
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


Vut11:
Sarawut

brcrista (approved):

joshmgross (approved):

joshmgross:
Thanks!

ci/python-publish.yml
```diff
@@ -21,11 +26,11 @@ jobs:
...
-        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
-        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-      run: |
-        python setup.py sdist bdist_wheel
-        twine upload dist/*
+        pip install build
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
```
webknjaz, line 33: @joshmgross FYI this is hugely outdated. The modern way is to use tokenless publishing via OIDC, that works out of the box with the current action version. This workflow keeps suggesting people to use a 3-years old commit.
See https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.
A similar problem existed with the GH docs that we got fixed almost a year ago and it's just pointing to the stable release in those examples: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-pypi#updating-your-github-actions-workflow.

965022-oss: Merge #2345

PR #914, opened 2021-05-26
Update push docker container to support container registry

pfleidi:
The changes in this PR update the `docker-publish` workflow in a few ways:

* It now supports logging in and pushing images to the [GH Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)
* Instead of using error-prone shell scripts to extract metadata and build/push images it's using Docker actions
* The registry being used is now defined in a variable

## Pre-requisites

  * I'm a GitHub employee so I'm assuming that doesn't apply to me 

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```

### DIFF
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -71,10 +71,6 @@ jobs:
         uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-          tags: |
-              type=ref,event=branch
-              type=ref,event=pr
-              type=semver,pattern={{version}}
 
       - name: Build and push Docker image
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -2,20 +2,22 @@ name: Docker
 
 on:
   push:
-    # Publish `$default-branch` as Docker `latest` image.
-    branches:
-      - $default-branch
-
+    branches: [ $default-branch ]
     # Publish `v1.2.3` tags as releases.
-    tags:
-      - v*
+    tags: [ 'v*.*.*' ]
+    pull_request_target:
+      branches: [ $default-branch ]
 
   # Run tests for any PRs.
   pull_request:
 
 env:
-  # TODO: Change variable to your image's name.
-  IMAGE_NAME: image
+  # TODO: Creates container named ACCOUNT/REPO
+  #       Append /IMAGE-NAME to create a container named ACCOUNT/REPO/IMAGE-NAME
+  IMAGE_NAME: ${{ github.repository }}
+
+  # Registry. Use docker.io for DockerHub
+  REGISTRY: ghcr.io
 
 jobs:
   # Run tests.
@@ -49,32 +51,31 @@ jobs:
       packages: write
 
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v2
 
-      - name: Build image
-        run: docker build . --file Dockerfile --tag $IMAGE_NAME
+      - name: Log into registry ${{ env.REGISTRY }}
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Log into registry
-        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+              type=ref,event=branch
+              type=ref,event=pr
+              type=semver,pattern={{version}}
 
-      - name: Push image
-        run: |
-          IMAGE_ID=docker.pkg.github.com/${{ github.repository }}/$IMAGE_NAME
-
-          # Change all uppercase to lowercase
-          IMAGE_ID=$(echo $IMAGE_ID | tr '[A-Z]' '[a-z]')
-
-          # Strip git ref prefix from version
-          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
-
-          # Strip "v" prefix from tag name
-          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
-
-          # Use Docker `latest` tag convention
-          [ "$VERSION" == "$default-branch" ] && VERSION=latest
-
-          echo IMAGE_ID=$IMAGE_ID
-          echo VERSION=$VERSION
-
-          docker tag $IMAGE_NAME $IMAGE_ID:$VERSION
-          docker push $IMAGE_ID:$VERSION
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          # Don't push on pull requests
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -60,7 +60,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Log into registry ${{ env.REGISTRY }}
-        uses: docker/login-action@v1
+        uses: docker/login-action@28218f9b04b4f3f62068d7b6ce6ca5b26e35336c
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.repository_owner }}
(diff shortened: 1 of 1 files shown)


ci/docker-publish.yml
```diff
@@ -49,32 +56,31 @@ jobs:
...
-          docker tag $IMAGE_NAME $IMAGE_ID:$VERSION
-          docker push $IMAGE_ID:$VERSION
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Log into registry ${{ env.REGISTRY }}
+        uses: docker/login-action@28218f9b04b4f3f62068d7b6ce6ca5b26e35336c
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.repository_owner }}
```
clarkbw, line 66: This doesn't actually matter because the service doesn't check it but don't we want `github.actor` here?

```suggestion
          username: ${{ github.actor }}
```

ci/docker-publish.yml
```diff
@@ -49,32 +56,31 @@ jobs:
...
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+              type=ref,event=branch
+              type=ref,event=pr
+              type=semver,pattern={{version}}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
```
clarkbw, line 82: If we can drop this and it defaults to `.` let's do that.

ci/docker-publish.yml
```diff
@@ -49,32 +56,31 @@ jobs:
...
+          # Don't push on pull requests
+          push: ${{ github.event_name != 'pull_request' }}
```
clarkbw, line 84: Unless this is the Dependabot issue do we need to prevent pushing from PRs?

ci/docker-publish.yml
```diff
@@ -49,32 +56,31 @@ jobs:
...
+
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+              type=ref,event=branch
+              type=ref,event=pr
+              type=semver,pattern={{version}}
```
clarkbw, line 77: It'd be best if we could get this down to zero elements. The `semver` seems required and hopefully we can change upstream to fix that.

ci/docker-publish.yml
pfleidi: I'll see if we can skip it and drop it if it does.

ci/docker-publish.yml
pfleidi: I was assuming that most developers would not want that because it would slow down their CI feedback circle. Depending on your CI pipeline, it could be pretty wasteful in terms of time and traffic to push every CI build to a registry so I assumed it was a sensible default.

In which scenarios would we want PRs to be pushed as well?

ci/docker-publish.yml
pfleidi: @clarkbw Do we want to track an issue somewhere to add the changes to upstream?

pfleidi: I've tried to slim it down to only the semver version but pushing to anything but a tag failed because the push operation needs a tag.

It failed with this message:

`Error: buildx call failed with: error: tag is needed when pushing to registry`

I feel like we need to take this upstream if we wanted to minimize the tags configuration.

pfleidi: Also: When removing the `event=pr` type, it technically works but we're getting warnings as well:

image

ci/docker-publish.yml
clarkbw: I don't think the push has a high cost in terms of time, if we've already built the container as part of our PR we might as well push it and any layers will be reused in a future build after merging.

ci/docker-publish.yml
crazy-max: @pfleidi It's safe to remove the `tags` input for the starter workflow. [Default entries if empty](https://github.com/docker/metadata-action#tags-input) should be enough.

ci/docker-publish.yml
```diff
@@ -49,32 +56,30 @@ jobs:
...
```
crazy-max, line 77: @pfleidi It's safe to remove the `tags` input for the starter workflow. [Default entries if empty](https://github.com/docker/metadata-action#tags-input) should be enough.

ci/docker-publish.yml
crazy-max: For GHCR it would be fine to push an image but would need to be done through `pull_request_target` event to be able to use the `GITHUB_TOKEN` secret right?

ci/docker-publish.yml
crazy-max: @clarkbw About that, it's true when using a GITHUB_TOKEN but not for a PAT right? I think GITHUB_TOKEN usage is the right move so do we need to also [update the example on docker-login action repo](https://github.com/docker/login-action#github-container-registry)?

ci/docker-publish.yml
clarkbw: Yes, though the odd thing about the `pull_request_target` is that it pulls the target branch so you're not actually building the PR branch. This makes the artifacts created somewhat useless.  Ex: If your PR branch `foo` is targeting `main` `pull_request_target` will do a build against `main` instead of `foo`; which isn't what we want. However if we use the `pull_request` and the user who created the PR doesn't have write access the push will fail because the `GITHUB_TOKEN` is read-only.

crazy-max: Ok so the condition on pull_request to push is correct here.

dominic59595:
pfleidi:packages-container-registry-workflow

dominic59595:
[pfleidi:packages-container-registry-workflow]() [](pfleidi:packages-container-registry-workflow) # #
Duplicate of #

dominic59595:
+
 on:
   push:
     branches: [ $default-branch ]

From 41a66c656e20c398aa7143b649d3678dae503afc Mon Sep 17 00:00:00 2001
From: Sven Pfleiderer <pfleidi@github.com>
Date: Wed, 26 May 2021 13:46:54 -0700
Subject: [PATCH 3/5] Use git shas rather than version tags

---
 ci/docker-publish.yml | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/ci/docker-publish.yml b/ci/docker-publish.yml
index 63b02f26e..26a5f89fa 100644
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -60,7 +60,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Log into registry ${{ env.REGISTRY }}
-        uses: docker/login-action@v1
+        uses: docker/login-action@28218f9b04b4f3f62068d7b6ce6ca5b26e35336c
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.repository_owner }}
@@ -68,7 +68,7 @@ jobs:
 
       - name: Extract Docker metadata
         id: meta
-        uses: docker/metadata-action@v3
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
@@ -77,7 +77,7 @@ jobs:
               type=semver,pattern={{version}}
 
       - name: Build and push Docker image
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
         with:
           context: .
           # Don't push on pull requests

From aa30f1448e6f181d695a784d284dbeaefc12a208 Mon Sep 17 00:00:00 2001
From: Sven Pfleiderer <sg+github@roothausen.de>
Date: Wed, 26 May 2021 14:07:29 -0700
Subject: [PATCH 4/5] Update ci/docker-publish.yml

Co-authored-by: Bryan Clark <clarkbw@github.com>
---
 ci/docker-publish.yml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/ci/docker-publish.yml b/ci/docker-publish.yml
index 26a5f89fa..d2672ab07 100644
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -63,7 +63,7 @@ jobs:
         uses: docker/login-action@28218f9b04b4f3f62068d7b6ce6ca5b26e35336c
         with:
           registry: ${{ env.REGISTRY }}
-          username: ${{ github.repository_owner }}
+          username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Extract Docker metadata

From 6fe53a9ebd963520a1da2b470d71380c14c284fb Mon Sep 17 00:00:00 2001
From: Sven Pfleiderer <pfleidi@github.com>
Date: Wed, 26 May 2021 15:03:41 -0700
Subject: [PATCH 5/5] Remove context configuration since it's already the
 default

---
 ci/docker-publish.yml | 1 -
 1 file changed, 1 deletion(-)

diff --git a/ci/docker-publish.yml b/ci/docker-publish.yml
index 26a5f89fa..c820f9177 100644
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -79,7 +79,6 @@ jobs:
       - name: Build and push Docker image
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
         with:
-          context: .
           # Don't push on pull requests
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}

ci/docker-publish.yml
pfleidi: I've removed that condition for now.

andymckay (approved):

PR #921, opened 2021-06-03
Simplify Docker publish workflow

crazy-max:
Follow-up #914 

Reduce and simplify to one job that can be used on any events.

cc @clarkbw @pfleidi

### DIFF
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -6,51 +6,26 @@ name: Docker
 # documentation.
 
 on:
+  schedule:
+    - cron: $cron-daily
   push:
     branches: [ $default-branch ]
-    # Publish `v1.2.3` tags as releases.
+    # Publish semver tags as releases.
     tags: [ 'v*.*.*' ]
-    pull_request_target:
-      branches: [ $default-branch ]
-
-  # Run tests for any PRs.
   pull_request:
+    branches: [ $default-branch ]
 
 env:
-  # TODO: Creates container named ACCOUNT/REPO
-  #       Append /IMAGE-NAME to create a container named ACCOUNT/REPO/IMAGE-NAME
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+  # github.repository as <account>/<repo>
   IMAGE_NAME: ${{ github.repository }}
 
-  # Registry. Use docker.io for DockerHub
-  REGISTRY: ghcr.io
 
 jobs:
-  # Run tests.
-  # See also https://docs.docker.com/docker-hub/builds/automated-testing/
-  test:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


ci/docker-publish.yml
```diff
@@ -6,51 +6,26 @@ name: Docker
 # documentation.
 
 on:
+  schedule:
+    - cron: '0 10 * * *' # everyday at 10am
```
crazy-max, line 10: Allow to produce nightly builds. Can be good to check upstream image (`FROM`).

ci/docker-publish.yml
```diff
@@ -6,51 +6,26 @@ name: Docker
...
   push:
     branches: [ $default-branch ]
-    # Publish `v1.2.3` tags as releases.
+    # Publish semver tags as releases.
```
crazy-max, line 13: Not necessary `v1.2.3`. Can be also be a beta release like `v2.0.8-beta.67` which is a valid semver but will not produce a latest tag because they are updated frequently, and contain many breaking changes that are (by the author's design) not yet fit for public consumption.

ci/docker-publish.yml
```diff
@@ -59,23 +34,29 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
+      # Login against a Docker registry except on PR
+      # https://github.com/docker/login-action
       - name: Log into registry ${{ env.REGISTRY }}
+        if: github.event_name != 'pull_request'
```
crazy-max, line 40: Do not login to the registry on PR

ci/docker-publish.yml
```diff
@@ -59,23 +34,30 @@ jobs:
...
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
+      # Build and push Docker image with Buildx (don't push on PR)
+      # https://github.com/docker/build-push-action
       - name: Build and push Docker image
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
         with:
-          # Don't push on pull requests
+          context: .
```
crazy-max, line 60: Use [path context](https://github.com/docker/build-push-action#path-context) here because the repo is checked out in the first step.

ci/docker-publish.yml
joshmgross: We try to avoid having starter workflows having a specific `cron` syntax set. I'd recommend using `$cron-daily` instead, which will choose a random time for this schedule to run daily when the starter workflow is selected.

```suggestion
    - cron: $cron-daily
```

crazy-max: @joshmgross Sure

pfleidi (approved):
Looks great. Thanks for doing this!

pfleidi (approved):

joshmgross (approved):
LGTM, thanks for reviewing @pfleidi

PR #995, opened 2021-07-08
Because That what we used for.

MA24th:
Check package structure and relevant paths.

### DIFF
--- a/ci/python-package.yml
+++ b/ci/python-package.yml
@@ -29,6 +29,8 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install flake8 pytest
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Install package
+      run: pip install -e .
     - name: Lint with flake8
       run: |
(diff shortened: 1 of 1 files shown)


andymckay:
👋🏾 This command for pip installs the current directory as an editable python package. Could you explain why you'd want to do that? This would be maybe useful for projects that want to become python packages, however less so for other projects. I'm not sure this command has value for a starter workflow at this point though.

CORONA19:
하이

MA24th:
> 👋🏾 This command for pip installs the current directory as an editable python package. Could you explain why you'd want to do that? This would be maybe useful for projects that want to become python packages, however less so for other projects. I'm not sure this command has value for a starter workflow at this point though.

If have multi package working in same project 
And these packages in test, this should make workflow more intensity.

andymckay:
I'd like to aim to keep the starter workflow simple and developing multiple packages in the same project seems like an unusual thing to be doing.

Thank you so much for your help, but I'd like to avoid adding complexity to this workflow.

PR #1006, opened 2021-07-15
fix: only run `deno fmt` and `deno lint` on one job

dsherret:
The `actions/checkout` action on windows checks out files with `\r\n` line endings instead of preserving them. This causes `deno fmt --check` to fail because the line endings are `\r\n` instead of `\n`.

Regardless, it is better to only run this on one job in order to save CI time. I've made the same change to the lint step.

cc @lucacasonato

### DIFF
--- a/ci/deno.yml
+++ b/ci/deno.yml
@@ -35,6 +35,7 @@ jobs:
 
       # Uncomment this step to verify the use of 'deno fmt' on each commit.
       # - name: Verify formatting
+      #   if: matrix.os == 'ubuntu-latest'
       #   run: deno fmt --check
 
       - name: Run linter
--- a/ci/deno.yml
+++ b/ci/deno.yml
@@ -39,6 +39,7 @@ jobs:
       #   run: deno fmt --check
 
       - name: Run linter
+        if: matrix.os == 'ubuntu-latest'
         run: deno lint
 
       - name: Cache dependencies


justinmchase:
It looks like a better solution would be to instead disable the autocrlf:

https://github.com/actions/checkout/issues/38#issuecomment-530970885

```yaml
    steps:
        # Add this new Step to disable it on all platforms, including Windows
      - name: Prepare git 
        run: git config --global core.autocrlf false 
        
      - name: Setup repo
        uses: actions/checkout@v2
        
      - name: Setup Deno
        uses: denoland/setup-deno@4a4e59637fa62bd6c086a216c7e4c5b457ea9e79
        with:
          deno-version: ${{ matrix.deno }}
```

dsherret:
@justinmchase this does have the advantage of saving CI time and I think something people would want to do anyway. Also, hopefully https://github.com/actions/checkout/issues/135 will eventually be resolved in actions/checkout.

justinmchase:
Actually you're right in my case, I only actually want ubuntu for now. Probably most people only care about one environment also. And if they need windows then can figure this trick out. 

So maybe it's better to have the `Prepare Git` step but leave it commented out with a comment about how it's necessary if you add the windows agent?

andymckay:
I think this gets back to the problem I mentioned [here](https://github.com/actions/starter-workflows/pull/604#pullrequestreview-474638763), in that a starter workflow should be as simple as possible and that users can add in the matrix if they want to. Because we've now gotten the matrix, we need to add stuff in to work around problems it causes - for example macOS minutes are expensive.

How would you feel about just removing the matrix?

justinmchase:
I think you're right.

champ219 (approved):

Lutthagone (approved):

dsherret:
@andymckay sounds good. Opened #1024.

PR #1007, opened 2021-07-15
Add Amazon CodeGuru Reviewer support

yueny2020:
## Pre-requisites

---

---

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```


code-scanning/codegurureviewer.yml
```diff
@@ -0,0 +1,55 @@
...
+
+  steps:
+    # Step 1: Checkout the repository and provide your AWS credentials
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
```
nickfyson, line 37: I'm afraid we need a full SHA here, rather than a tag.

code-scanning/codegurureviewer.yml
```diff
@@ -0,0 +1,55 @@
...
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-region: us-west-2  # Region to access CodeGuru
+
+    # Step 2: Add CodeGuru Reviewer Action
+    - name: CodeGuru Reviewer
+      uses: aws-actions/codeguru-reviewer@v1.1
```
nickfyson, line 45: I'm afraid we need a full SHA here, rather than a tag.

nickfyson (changes requested):
Thanks for this, and it's looking great! 👍



Just a couple of places where we need a full SHA rather than a tag, then it's good to go! 🙂

code-scanning/codegurureviewer.yml
yueny2020: This step is to use a GitHub action, same as the usage of `uses: actions/checkout@v2`. 
I don't think I can provide a SHA here.

code-scanning/codegurureviewer.yml
yueny2020: If I put a full SHA here, does it mean every time my action has an update even the version keeps same, I also need to raise a pr to update the SHA in this repo?

Or could you please elaborate why we need a full SHA here not a tag?

nickfyson: I'm afraid it's a requirement we've settled on from a security perspective. When we are providing templates that use actions which are not under our control then a full SHA better ensures users get the expected version.

This does mean updates to the action will requires a small PR here in order to be reflected in the starter workflows page. This shouldn't be any more burdensome than bumping eg `v1.1` to `v1.2`, but apologies that the extra admin is necessary. (We hope that users will find Dependabot useful for keeping their workflows up to date once in use!)

Do let me know if you have further questions! 🙂

code-scanning/codegurureviewer.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-region: us-west-2  # Region to access CodeGuru
+
+    # Step 2: Add CodeGuru Reviewer Action
+    - name: CodeGuru Reviewer
+      uses: aws-actions/codeguru-reviewer@e96af14e3ba9bac3cca62a524bd5a8345038c684
```
yueny2020, line 45: Replaced the version with the full SHA.

code-scanning/codegurureviewer.yml
nickfyson: Sorry, but we still need a full hash here as well! Should point to the same commit, just with a SHA rather than a tag! 🙂

yueny2020: Solved.

code-scanning/codegurureviewer.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         fetch-depth: 0
 
     - name: Configure AWS Credentials
-      uses: aws-actions/configure-aws-credentials@v1
+      uses: aws-actions/configure-aws-credentials@0d9a5be0dceea74e09396820e1e522ba4a110d2f
```
yueny2020, line 37: Resolved.

code-scanning/codegurureviewer.yml
```diff
@@ -0,0 +1,55 @@
...
+    # Step 2: Add CodeGuru Reviewer Action
+    - name: CodeGuru Reviewer
+      uses: aws-actions/codeguru-reviewer@e96af14e3ba9bac3cca62a524bd5a8345038c684
+      with:
+        build_path: target # build artifact(s) directory
+        s3_bucket: codeguru-reviewer-my-bucket  # S3 Bucket with "codeguru-reviewer-*" prefix
+
+    # Step 3: Upload results into GitHub
+    - name: Upload review result
+      if: ${{ github.event_name != 'push' }}
```
nickfyson, line 52:
```suggestion
```

Is there a reason why this limitation is included? Code scanning relies on results being uploaded for `pull_request` events in order to provide results on PRs.

nickfyson (changes requested):
My apologies for only just noticing this, and for delaying the PR further!

code-scanning/codegurureviewer.yml
aschembo-dev: This was added because for `push` events, our code scanner only runs a scan on the changed lines of code. Thus, some recommendations that were populated in the security tab from a scheduled scan would get closed when a push scan was run (as the new SARIF file would only contain recommendations on the changed lines). 

We do still upload the results for `pull_request` events, this is just excluding `push` events. Example seen here: https://github.com/aschembo-dev/testcicd/pull/158

nickfyson: Sorry, yes, it was late and I got that backwards. 😄

We do rely on both `push` and `pull_request` analyses to provide results on PRs though, letting us determine which alerts have been introduced in that PR (relative to the baseline analysis).

Is your tool already accounting for that in some way?

aschembo-dev: The issue is that a `push` event scan will mark existing recommendations (in the security tab) as closed even if they may not have been fixed due to our `push` scan having recommendations only on the changed lines of code. 

I can understand that for new alerts to be added in a PR, there needs to be a `push` event then, but we are already providing these recommendations in the comment of the PR, and the `push` event could close existing ones. 

This is the reason we disabled `push` events from uploading SARIF file.

yueny2020:
Hi @nickfyson,

What else do we need to do to merge this pr?

nickfyson:
@yueny2020 Sorry for the delay in responding, but I believe there have now been direct discussions between our teams. Thanks for the PR, even though it didn't work out on this occasion!

PR #1012, opened 2021-07-18
ci(workflow): add cache to workflows using actions/setup-node

oscard0m:
## Description

Add `cache` to workflows using `actions/setup-node`

## Context

`setup-node` GitHub Action just released a new option to add cache to steps using it.

You can find the details here: https://github.blog/changelog/2021-07-02-github-actions-setup-node-now-supports-dependency-caching/

---

🤖 This PR has been generated automatically by [this octoherd script](https://github.com/oscard0m/octoherd-script-add-cache-to-node-github-action), feel free to run it in your GitHub user/org repositories! 💪🏾

### DIFF
--- a/.github/workflows/validate-data.yaml
+++ b/.github/workflows/validate-data.yaml
@@ -1,9 +1,8 @@
 name: Validate Data
 
 on:
-  push:
-  pull_request:
-
+  push: null
+  pull_request: null
 jobs:
   validate-data:
     runs-on: ubuntu-latest
@@ -13,6 +12,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "12"
+          cache: npm
 
       - name: Validate workflows
         run: |
--- a/.github/workflows/validate-data.yaml
+++ b/.github/workflows/validate-data.yaml
@@ -1,8 +1,8 @@
 name: Validate Data
 
 on:
-  push: null
-  pull_request: null
+  push:
+  pull_request:
 jobs:
   validate-data:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


Imranba23 (approved):

Imranba23 (approved):

eliasbuchwald (approved):

.github/workflows/validate-data.yaml
```diff
@@ -1,9 +1,8 @@
 name: Validate Data
 
 on:
-  push:
-  pull_request:
-
+  push: null
+  pull_request: null
```
andymckay, line 5:
```suggestion
  push:
  pull_request:
```

The `null` isn't needed and isn't in other starter workflows, or in our docs so let's remove that.

andymckay (commented):
With the suggested changes, this looks good to go, thank you.

017963355 (approved):

andymckay (approved):
Thank you @oscard0m

.github/workflows/validate-data.yaml
```diff
@@ -13,6 +12,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "12"
+          cache: npm
```
andymckay, line 15: And similarly here, this won't work.

andymckay:
I'll note that the starter workflow script our users get is [here](https://github.com/actions/starter-workflows/blob/7135450ec14e7e47bf3e41dd65d399e5e315fa39/ci/node.js.yml) and has the `cache` value set. So I don't think we want this PR at all at this point, sorry.

oscard0m:
> I'll note that the starter workflow script our users get is [here](https://github.com/actions/starter-workflows/blob/7135450ec14e7e47bf3e41dd65d399e5e315fa39/ci/node.js.yml) and has the `cache` value set. So I don't think we want this PR at all at this point, sorry.

Thanks for the feedback and your time reviewing this

PR #1024, opened 2021-07-26
Remove matrix for Deno template

dsherret:
Based on #1006, this removes the matrix to simplify the template and fix the deno fmt issue on Windows with actions/checkout as outlined in #1006.

> The actions/checkout action on windows checks out files with \r\n line endings instead of preserving them. This causes deno fmt --check to fail because the line endings are \r\n instead of \n.

### DIFF
--- a/ci/deno.yml
+++ b/ci/deno.yml
@@ -3,7 +3,7 @@
 # separate terms of service, privacy policy, and support
 # documentation.
 
-# This workflow will install Deno and run tests across stable and canary builds on Windows, Ubuntu and macOS.
+# This workflow will install Deno then run Deno lint and test.
 # For more information see: https://github.com/denoland/setup-deno
 
 name: Deno
@@ -16,12 +16,7 @@ on:
 
 jobs:
   test:
-    runs-on: ${{ matrix.os }} # runs a test on Ubuntu, Windows and macOS
-
-    strategy:
-      matrix:
-        deno: ["v1.x", "canary"]
-        os: [macOS-latest, windows-latest, ubuntu-latest]
+    runs-on: ubuntu-latest
 
     steps:
       - name: Setup repo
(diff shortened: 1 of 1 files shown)


ci/deno.yml
```diff
@@ -16,22 +16,17 @@ on:
...
+    runs-on: ubuntu-latest
 
     steps:
       - name: Setup repo
         uses: actions/checkout@v2
 
       - name: Setup Deno
         # uses: denoland/setup-deno@v1
-        uses: denoland/setup-deno@4a4e59637fa62bd6c086a216c7e4c5b457ea9e79
+        uses: denoland/setup-deno@004814556e37c54a2f6e31384c9e18e9833173669
```
dsherret, line 27: Not sure why this is done, but ugraded to https://github.com/denoland/setup-deno/releases/tag/v1.1.0

ci/deno.yml
```diff
@@ -16,22 +16,17 @@ on:
 
 jobs:
   test:
-    runs-on: ${{ matrix.os }} # runs a test on Ubuntu, Windows and macOS
-
-    strategy:
-      matrix:
-        deno: ["v1.x", "canary"]
```
joshmgross, line 23: Should we keep the deno version matrix and only remove the OS matrix?

justinmchase: This is only a quick start too, people can adapt their own pipelines based on their needs after getting kick started with this. I agree its nice to have the simplest, fasted and most stable setup for the initial pipeline template.

justinmchase (approved):

joshmgross (approved):
LGTM, thanks!

Vut11:
ช่วยกู้เงินได้หน่อย

Vut11:
ช่วยกู้เงินได้หน่อย

PR #1040, opened 2021-08-12
add specific permissions to the labeler-triage workflow

nickfyson:
This PR adds specific permissions to the existing `labeler-triage` workflow

### DIFF
--- a/.github/workflows/labeler-triage.yml
+++ b/.github/workflows/labeler-triage.yml
@@ -1,4 +1,9 @@
 name: "Pull Request Labeler"
+
+permissions:
+  contents: read
+  pull-requests: write
+
 on:
 - pull_request_target
 


.github/workflows/labeler-triage.yml
```diff
@@ -1,4 +1,9 @@
 name: "Pull Request Labeler"
+
+permissions:
+  contents: read
+  pull-requests: write
```
anaarmas, line 5: do these govern what scopes the `secrets.GITHUB_TOKEN` has?

nickfyson: Yep, that's right. The current default gives access to a lot more, but I believe it is now best practice to explicitly grant exactly what is required.

anaarmas: This looks great ❤️

anaarmas (approved):

PR #1044, opened 2021-08-16
Properties.json - name and icon validation

aparna-ravindra:
Adding extra validation to the properties.json file:
1. `name` field should not be empty
2. `name` should be unique within the repo
3. `iconName` - if octicon is used, then octicon icon name should be specified.


script/validate-data/index.ts
```diff
@@ -77,17 +87,28 @@ async function checkWorkflow(workflowPath: string, propertiesPath: string): Prom
...
     workflowErrors.errors = res.errors.map(e => e.toString())
-
-    if (properties.iconName && !properties.iconName.startsWith("octicon")) {
-      try {
-        await fs.access(`../../icons/${properties.iconName}.svg`)
-      } catch (e) {
-        workflowErrors.errors.push(`No icon named ${properties.iconName} found`)
+    
+    if (properties.iconName) {
+      if(!properties.iconName.startsWith("octicon ")) {
```
ashwinsangem, line 98: Should we use regex to handle one or more spaces after `octicon`?
and should we handle it with lower casing, like `properties.iconName.tolower.startsWith`?

aparna-ravindra: - Used regex for checking the string.
- The value is case sensitive as seen [here](https://github.com/github/github/blob/fb08a1cd44254fa796fac532afb1f9d55c9e82f1/app/views/repository_actions/_workflow_template.html.erb#L44)

script/validate-data/index.ts
```diff
@@ -41,7 +42,7 @@ const propertiesSchema = {
 
 async function checkWorkflows(folders: string[]): Promise<WorkflowWithErrors[]> {
   const result: WorkflowWithErrors[] = []
-
+  const workflow_template_names : string[] = []
```
ashwinsangem, line 45: Since this is going to be a unique, you can use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)

ashwinsangem: https://stackoverflow.com/questions/40180641/hashset-in-typescript

script/validate-data/index.ts
```diff
@@ -54,9 +55,17 @@ async function checkWorkflows(folders: string[]): Promise<WorkflowWithErrors[]>
...
-        if (errors.errors.length > 0) {
-          result.push(errors)
+        const workflowWithErrors = await checkWorkflow(workflowFilePath, propertiesFilePath);
+        if(workflowWithErrors.name) {
+          if(workflow_template_names.includes(workflowWithErrors.name)) {
+            workflowWithErrors.errors.push(`Workflow template name "${workflowWithErrors.name}" already exists`)
+          }
+          else {
+            workflow_template_names.push(workflowWithErrors.name)
+          }
```
ashwinsangem, line 65: I guess Set will simplify this code.

script/validate-data/index.ts
```diff
@@ -77,17 +87,28 @@ async function checkWorkflow(workflowPath: string, propertiesPath: string): Prom
...
     workflowErrors.errors = res.errors.map(e => e.toString())
-
-    if (properties.iconName && !properties.iconName.startsWith("octicon")) {
-      try {
-        await fs.access(`../../icons/${properties.iconName}.svg`)
-      } catch (e) {
-        workflowErrors.errors.push(`No icon named ${properties.iconName} found`)
+    
+    if (properties.iconName) {
+      if(! /^octicon\s+/.test(properties.iconName)) {
```
ashwinsangem, line 98: I guess you can club both the above conditions into a single `if`.

script/validate-data/index.ts
ashwinsangem: > const mySet1 = new Set()
> 
> mySet1.add(1)           // Set [ 1 ]
> mySet1.add(5)        // Set [ 1, 5 ]
> mySet1.add(5)           // Set [ 1, 5 ]

Looking at the above,

`if(mySet1.size == mySet1.add(5).size) {workflowWithErrors.errors.push(`Workflow template name "${workflowWithErrors.name}" already exists`) }`

ashwinsangem (approved):
LGTM.

PR #1067, opened 2021-08-30
Validate categories in properties.json against Linguist/Scout and comment on PR

aparna-ravindra:
The categories listed in the _.properties.json_ file of a template determine the section under which the template appears in the [actions/new](https://github.com/github/github/actions/new) page. Also, it plays an important role in the "Workflows made for your repository" section of the page. The recommendations made to the user depends on the languages and tech_stacks identified by [Linguist](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) and [Scout](https://github.com/github/scout/blob/main/lib/scout/tech_stacks.yml).

Therefore, capturing any typos or unrecognised categories at the time of template creation will help improving the recommendation section. 

This PR aims at validating the `categories` field present in the _.properties.json_ file for each of the templates and bubbling up any category that is not recognised by either Linguist or Scout. The aim is to make the reviewer aware of such cases by commenting on the PR. The reviewer can then choose to recommend the right set of categories for the template. 

Points of interest:
- Other than the languages and tech_stacks, the `categories` field may also contain the keywords such as "Automation", "Deployment", "Code Scanning", etc which are used to determine the section they appear in the actions/new page. Hence a list of such keywords is maintained in the `settings` json and used to suppress false errors. 
- At the time of writing this, Linguist is a private repository with unpublished gem. Therefore, a manually built version of Scout is added to the repository and is installed when the workflow runs. 
- We need to come up with a plan to maintain the Scout version used in the script. 
- The workflow runs on "pull_request_target" similar to [labler](https://github.com/actions/starter-workflows/blob/main/.github/workflows/labeler-triage.yml). I did come across the doc which says "pull_request_target" is not very safe. Linking the doc [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). In gist the doc says 
> combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.

At the same time it mentions:

> If your workflow scenario simply requires commenting on the PR, but does not require a check out of the modified code, using pull_request_target is a logical shortcut.

Screenshot of the expected output from the PR:

image


aligetget (commented):
.github/workflows/validate-categories.yaml

mervyd (commented):
#1599 #1600

PR #1068, opened 2021-08-30
Add a descriptive comment into stale.yml

duduribeiro:
This PR adds a descriptive comment int "stale.yml" so user know what this does and how adjust.

 because user's can come to this workflow as a template
directly from their issue page and this extra 
This PR adds a descriptive comment int "stale.yml" so user know what this does and how adjust.

This was a suggestion from @Carmel-S and @jenschelkopf and we think this can be helpful because users can come to this workflow as a template directly from their issue page and this extra content will help them understand what this is.

image

### DIFF
--- a/automation/stale.yml
+++ b/automation/stale.yml
@@ -1,3 +1,8 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
 name: Mark stale issues and pull requests
 
 on:


Carmel-S:
The stale issue workflow template will now include this copy "This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time."at the top of the file to ensure the YAML file is related to the button they clicked on the issues screen. 

Screen Shot 2021-08-30 at 1 06 54 PM

andymckay (approved):

andymckay:
Thanks @duduribeiro

PR #1073, opened 2021-09-03
Ruby on Rails with Postgres - Lint and Test

aparna-ravindra:
This workflow will download a prebuilt Ruby version, install dependencies , run linters and run tests with Rake.
It sets up a Postgres database for running tests. The inputs for database setup is to be provided in the `env` map. The password is taken from the repository secret `RAILS_TEST_DB_PASSWORD`.

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```


ci/rails-postgres.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Run security checks
+        # Add or Replace any other security checks here
+        run: |
+          bin/bundler-audit --update
+          bin/brakeman -q -w2
+      - name: Run linters
+        # Add or Replace any other Linters here
+        run: |
+          bin/rubocop --parallel
+  run-tests:
```
ashwinsangem, line 53: We should split lint and test jobs into 2 ymls. I see 2 reasons:
I) I see that the CI templates in this repo are modular and deal with mostly one job and so there are templates for each step in CI for a given language.
II) The issue with multiple jobs in a single workflow is that if one job fails out of 2 or more jobs in your workflow, then you’ll have to re-run all those jobs for the workflow status to be a success.

(OR)
I see that 5 steps are common across the 2 jobs. So, you can instead make it into a single job and make this file small. So, this way also, if linting fails, further steps are not run.

ci/rails-postgres.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: cache
+        uses: actions/cache@v1
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: Install packages
+        run: |
+          yarn install --pure-lockfile
```
bishal-pdMSFT, line 43: Is yarn more popular than bundler? Why not just do bundler install?

ci/rails-postgres.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Find yarn cache location
+        id: yarn-cache
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: cache
+        uses: actions/cache@v1
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
```
bishal-pdMSFT, line 40: Is there a new recommendation to use caching? I have not seem the caching being used in node templates for example.

PR #1078, opened 2021-09-07
Build Rails and run linters

aparna-ravindra:
This workflow will download a prebuilt Ruby version, install dependencies and run linters.
## Tasks

**Some general notes:**

    ```
    # documentation.
    ```

### DIFF
--- a/ci/properties/rails-lint.properties.json
+++ b/ci/properties/rails-lint.properties.json
@@ -1,0 +1,6 @@
+{
+    "name": "Rails - Build and Run Linters",
+    "description": "Build Rails application and run linters",
+    "iconName": "ruby",
+    "categories": ["Ruby", "Rails"]
+}
--- a/ci/rails-lint.yml
+++ b/ci/rails-lint.yml
@@ -1,0 +1,32 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will download a prebuilt Ruby version, install dependencies, and  run linters
+name: Build Rails and run linters
+
+on:
+  push:
+    branches: [ $default-branch ]
+  pull_request:
+    branches: [ $default-branch ]
+jobs:
+  run-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup Ruby and install gems
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      # Add or Replace any other security checks here
+      - name: Run security checks
+        run: |
+          bin/bundler-audit --update
+          bin/brakeman -q -w2
+      # Add or Replace any other Linters here
+      - name: Run linters
+        run: |
+          bin/rubocop --parallel


ci/properties/rails-lint.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Rails - Build and Run Linter",
```
ashwinsangem, line 2: You can use `linters`(plural) everywhere to be consistent.

ci/rails-lint.yml
```diff
@@ -0,0 +1,32 @@
...
+    branches: [ $default-branch ]
+jobs:
+  run-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup Ruby and install gems
+        uses: ruby/setup-ruby@v1
```
bishal-pdMSFT, line 21: This will install gems as well? Which file will it use for gemfile? Do we need to detect that?

ci/rails-lint.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          bundler-cache: true
```
bishal-pdMSFT, line 23: What does this do?

aparna-ravindra: this runs 'bundle install' and caches installed gems automatically.
Doc: https://github.com/ruby/setup-ruby

ci/rails-lint.yml
aparna-ravindra: It picks Gemfile from the current working directory.

ci/properties/rails-lint.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Rails - Build and Run Linters",
+    "description": "Build Rails application and run linters",
+    "iconName": "ruby",
+    "categories": ["Ruby", "Rails"]
```
NinadKavimandan, line 5: we are not detecting Rails in scout yet, so make sure to follow this nomenclature once we add support in scout

ci/rails-lint.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull_request:
+    branches: [ $default-branch ]
+jobs:
+  run-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup Ruby and install gems
```
NinadKavimandan, line 20: In the existing ruby template - https://github.com/actions/starter-workflows/blob/main/ci/ruby.yml, we are using a matrix of different ruby versions. Should we do the same?

aparna-ravindra: I was wondering that. But if the project mentions a particular Ruby version in the Gem file, then the workflow will fail with errors like:
`Your Ruby version is 3.0.2, but your Gemfile specified 2.7.2`. Hence chose not to use version matrix strategy.

NinadKavimandan:
That's a lot of commits @aparna-ravindra, can we reduce it a bit?

ci/rails-lint.yml
NinadKavimandan: Fair enough

NinadKavimandan (approved):
LGTM 👍

Serge2703 (approved):

Serge2703 (changes requested):
ci/rails-lint.yml

Serge2703:
Testing templates

Serge2703 (commented):
Bundler-cache/ruby

Serge2703 (commented):
> `
Duplicate of #`

Serge2703 (approved):
ci/rails-lint.yml

bishal-pdMSFT:
> That's a lot of commits @aparna-ravindra, can we reduce it a bit?

Best is to do the squash commit, it will take only one commit to master

ci/properties/rails-lint.properties.json
bishal-pdMSFT: @NinadKavimandan @aparna-ravindra wouldn’t it be better to add Rails detection to Scout now itself? Or are you planning that at a later point in time?

bishal-pdMSFT (approved):

ci/properties/rails-lint.properties.json
aparna-ravindra: Yes, I am working on it. Will raise a PR asap.

ci/rails-lint.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will download a prebuilt Ruby version, install dependencies, and  run linters
+name: Build Rails and run linters
```
joshmgross, line 6: 👋  There's no "Building" for Rails, should this just be "Run linters"?

joshmgross (commented):
For security reasons, we shouldn't pin third party actions to tags or branches. It's preferred to use full commit SHAs, for example:
https://github.com/actions/starter-workflows/blob/29e8b6c38a618f6fe6daf8863d0d41c1ff40f132/ci/ruby.yml#L27-L30

ci/rails-lint.yml
aparna-ravindra: @joshmgross Thanks for the inputs. I will raise another PR to incorporate these.

aparna-ravindra: @joshmgross Addressed your inputs in this PR -> https://github.com/actions/starter-workflows/pull/1086

PR #1095, opened 2021-09-15
Adding template - Build Xcode project

aparna-ravindra:
The existing [ios template](https://github.com/actions/starter-workflows/blob/238e55b9b4252387b55b4dbcbb2fd71258ac6e0f/ci/ios.yml#L37) is specific to iOS platform. The template added in this PR is more generic. 

This template has "Objective-C" in its list of categories in the .properties.json file. Therefore, it will be recommended to repositories which contain Objective-C code. This is with the understanding that the most common usecase of Objective-C is projects on Apple platform which are most likely build using Xcode. 

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```


ci/properties/objective-c-xcode.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Build Xcode project",
```
bishal-pdMSFT, line 2: Will this template build any xcode project and not just Objective-c language ones? Is there a case where it can't build objective-c code?

bishal-pdMSFT, line 2: Change to a more succinct name - `Xcode project`

bishal-pdMSFT: @aparna-ravindra my question was other way round. Is it possible that an Objective-c code written using xcode does not build using the above command? If so, then adding `objective-c` in categories might be questionable.

aparna-ravindra: No, the command will build all objective-c projects developed using xcode.

bishal-pdMSFT: got it. Thanks

bishal-pdMSFT: Are you planning to take this comment?

bishal-pdMSFT (approved):

bishal-pdMSFT:
LGTM!

PR #1099, opened 2021-09-16
Lint - Categories in the properties.json file 

aparna-ravindra:
### Context
The categories listed in the _.properties.json_ file of a template determine the section under which the template appears in the [actions/new](https://github.com/github/github/actions/new) page. 

### What does this PR contain?
This PR aims at validating the `categories` field present in the _.properties.json_ file for each of the templates and identifying that it contains at least one of the supported categories. The aim is to make the reviewer aware of such cases by commenting on the PR. The reviewer can then choose to recommend the right category for the template. 

The list of recognised categories is maintained in the settings.json file and is shared in the comment with the PR owner.

This PR contains two workflows
1. Checkout the PR branch and validate categories for all the templates. Save the result in a workflow artifact.
2. Download the workflow artifact and comment on PR.

The comment on the PR appears like this: https://github.com/aparna-ravindra/starter-workflows/pull/42#issuecomment-920592073

tldr;
Two workflows were needed because the workflows running on the PR branch(from forked repo) do not have permission to comment on the issue. This has been documented [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). In gist the doc says 
> combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.

At the same time it mentions:

> If your workflow scenario simply requires commenting on the PR, but does not require a check out of the modified code, using pull_request_target is a logical shortcut.

Therefore the approach described in the doc has been followed.


script/validate-categories/validate.rb
```diff
@@ -0,0 +1,37 @@
+require 'json'
```
bishal-pdMSFT, line 1: Why not use validate-data script itself?

aparna-ravindra: The reasons I plan on keeping the script separate
1. Modularity : The validate-data script already does multiple checks. We expect to add more checks in the validate-categories script, which will make validate-data bulky if combined.
2. Validate-Data is a mandatory check on PR, failing which will block PR merge. We intend Validate-categories to nudge the PR author with information on best practices for the values of the category field. There is no plan on failing the check.

script/validate-categories/validate.rb
```diff
@@ -0,0 +1,37 @@
...
+
+settings = JSON.parse(File.read('./settings.json'))
+folders = settings['folders']
+@allowed_categories = settings['allowed_categories']
+
+def validateCategories(categories)
+    return categories.nil? || (categories.is_a?(Array) && ! categories.select{|l| @allowed_categories.detect{|p| p.casecmp(l) == 0 } }.empty? )
```
bishal-pdMSFT, line 8: Will array [intersection](https://blog.saeloun.com/2019/10/14/ruby-adds-array-intersection.html#arrayintersection) achieve same result as `select` and `detect`?

bishal-pdMSFT, line 8: If categories is nil, then validate should fail

script/validate-categories/validate.rb
```diff
@@ -0,0 +1,37 @@
...
+end
+
+result = []
+for folder in folders
+    files = Dir.entries(folder).select {|entry| File.file?(File.join(folder, entry)) && (File.extname(entry) == ".yaml" || File.extname(entry) == ".yml") }
```
bishal-pdMSFT, line 13: Do we need to co-relate with workflow yml file? Or just simply look through all properties file?

.github/workflows/validate-category-on-pr.yaml
```diff
@@ -0,0 +1,62 @@
+name: Validate Categories on PR
+
+# read-only repo token
+# no access to secrets
+on:
+  pull_request:
```
bishal-pdMSFT, line 6: add on `push` event as well. `ghes` branch has direct pushes.

.github/workflows/validate-category-on-pr.yaml
```diff
@@ -0,0 +1,62 @@
+name: Validate Categories on PR
```
bishal-pdMSFT, line 1: Please add a comment on what are the valid categories

script/validate-categories/validate.rb
bishal-pdMSFT: Ah! I think I get it now. category validation is an optional thing and we only put a PR message. Seems like we discussed it long back.

script/validate-categories/validate.rb
aparna-ravindra: Need this to be case-insensitive

aparna-ravindra: categories is an optional field. We do have a blank template today where there are no categories.

script/validate-categories/validate.rb
aparna-ravindra: The co-relation is taken care by the validate-data script.

.github/workflows/validate-category-on-pr.yaml
aparna-ravindra: ghes branch is updated only through the sync-ghes workflow, I believe. 
Also, we require a PR to comment if needed. Otherwise, the validation script is run, but the result is not shared anywhere.

bishal-pdMSFT: Got it. So, essentially master branch is the gatekeeper for ensuring correct categories.

script/validate-categories/validate.rb
bishal-pdMSFT: No, I meant why do we need to first find the workflow file name and then look for properties.json for that? We can directly iterate through all properties.json and validate category.

script/validate-categories/validate.rb
bishal-pdMSFT: @aparna-ravindra actually just re-thinking about it, currently we are only validating the template type category and not tech stack category. And if so, then we should fail the validation just like validate-data. And hence we can make this part of validate-data itself. Thoughts?

script/validate-categories/validate.rb
bishal-pdMSFT: This seems troublesome to me. Blank is a special case we should handle is separately. For any other template, we can disallow missing `categories`.

fatemehramezanli: من میخوام ارزهایی که در کیف پول تراست دارم را انتقال بدم به صرافی وبفروشم .باید چه کاری انجام دهم؟

.github/workflows/validate-category-on-pr.yaml
```diff
@@ -0,0 +1,63 @@
...
+          MY_STRING="${MY_STRING//$'\n'/'%0A'}"
+          MY_STRING="${MY_STRING//$'\r'/'%0D'}"
+          MY_STRING="${MY_STRING}%0A${{join(steps.validate-categories.outputs.*, '%0A')}}"
+          echo "::set-output name=content::$MY_STRING"
+          
+
+      - name: Save Validation Output 
+        if: ${{steps.comment-format.outputs.content && steps.comment-format.outputs.content != '' }}
+        run: |
+          mkdir -p ./category_validation
```
bishal-pdMSFT, line 55: This approach for publishing artifact is needed due to the reason that PR workflow won't have permission to write to PR in case of forked repos most probably. Isn't it?
Would it be easier to write a bot or as I mentioned in another comment to make category validation part of validate-data

PR #1127, opened 2021-09-27
Removing "deployment" templates from sync-ghes

aparna-ravindra:


script/sync-ghes/settings.json
```diff
@@ -2,7 +2,6 @@
   "folders": [
     "../../ci",
     "../../automation",
-    "../../deployments",
```
ashwinsangem, line 5: Do we need the `partners` array as we're removing this entry as that array might not get used?

aparna-ravindra: Yes, we need it. It is being used in the script.

ashwinsangem (approved):

Zarghampor:
[]()

PR #1129, opened 2021-09-27
start validating code-scanning workflows

nickfyson:
This PR explicitly enables the validation for Code Scanning workflows.

👋 @aparna-ravindra Would you mind taking a look at this? I see you've been active with these scripts, so are probably well placed to review! 🙂


script/validate-data/index.ts
```diff
@@ -49,7 +49,7 @@ async function checkWorkflows(folders: string[]): Promise<WorkflowWithErrors[]>
     });
 
     for (const e of dir) {
-      if (e.isFile()) {
+      if (e.isFile() && extname(e.name) === ".yml") {
```
aparna-ravindra, line 52: This looks good to me. Would you like to check for both `.yml` and `.yaml` here?

nickfyson: 👍 will do!

anaarmas (approved):
approving based on previous approval + diff since

itepopup11:
nukadu

itepopup11:
> This PR explicitly enables the validation for Code Scanning workflows.
> 
> 👋 @aparna-ravindra Would you mind taking a look at this? I see you've been active with these scripts, so are probably well placed to review! 🙂

PR #1162, opened 2021-10-15
Add partner templates for Azure Web Apps

JasonFreeberg:
- Renames the `azure.yml` template to a Node-specific name
- Adds templates for Java, Python, .NET Core, PHP, and Container apps

Todo:
- [x] Get review from Pulkit and team
- [x] Add setup notes (secrets, publish profile)
- [x] Add workflow for PHP
- [x] Add workflow for .NET


deployments/azure-webapps-container.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Log in to GitHub container registry
+        uses: docker/login-action@v1
```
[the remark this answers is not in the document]
skedwards88: From the docs point of view (as I look into adding guides for these additional templates), even if you use a version number, we will translate it to a SHA in the docs version. (This is our practice for all third party actions to demonstrate good security practices.)

deployments/azure-webapps-java-jar.yml
```diff
@@ -0,0 +1,72 @@
+# This workflow will build and push a node.js application to an Azure Web App when a commit is pushed to your default branch.
+#
+# This workflow assumes you have already created the target Azure App Service web app.
+# For instructions see https://docs.microsoft.com/en-us/azure/app-service/quickstart-nodejs?tabs=linux&pivots=development-environment-cli
```
skedwards88, line 4: should this not be `node`?

deployments/azure-webapps-php.yml
```diff
@@ -0,0 +1,80 @@
+# This workflow will build and push a node.js application to an Azure Web App when a commit is pushed to your default branch.
+#
+# This workflow assumes you have already created the target Azure App Service web app.
+# For instructions see https://docs.microsoft.com/en-us/azure/app-service/quickstart-nodejs?tabs=linux&pivots=development-environment-cli
```
skedwards88, line 4: Same here--should this reference PHP instead of Node?

deployments/azure-webapps-python.yml
```diff
@@ -0,0 +1,81 @@
+# This workflow will build and push a node.js application to an Azure Web App when a commit is pushed to your default branch.
+#
+# This workflow assumes you have already created the target Azure App Service web app.
+# For instructions see https://docs.microsoft.com/en-us/azure/app-service/quickstart-nodejs?tabs=linux&pivots=development-environment-cli
```
skedwards88, line 4: Here as well to reference Python instead of Node.js.

deployments/azure-webapps-node.yml
```diff
@@ -0,0 +1,73 @@
...
+      run: |
+        npm install
+        npm run build --if-present
+        npm run test --if-present
+
+    - name: Upload artifact for deployment job
+      uses: actions/upload-artifact@v2
+      with:
+        name: node-app
+        path: .
```
skedwards88, line 52: So that I can communicate better in the docs, can you share your reasoning for splitting building and deploying into two jobs, and uploading/downloading the artifact (as opposed to the current workflow, which builds and deploys in a single job)?

deployments/azure-webapps-java-jar.yml
jeremyepling: The workflow is for Java so the comments shouldn't be about node.

deployments/azure-webapps-java-jar.yml
```diff
@@ -0,0 +1,72 @@
...
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Java version
+        uses: actions/setup-java@v2.3.1
+        with:
+          java-version: ${{ env.JAVA_VERSION }}
```
jeremyepling, line 42: We should enable caching.

deployments/azure-webapps-java-jar.yml
JasonFreeberg: Thanks, copy/paste error. Fixed it

deployments/azure-webapps-php.yml
JasonFreeberg: Thanks, copy/paste error. Fixed it

deployments/azure-webapps-python.yml
JasonFreeberg: Thanks, copy/paste error. Fixed it

deployments/azure-webapps-container.yml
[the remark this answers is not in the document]
JasonFreeberg: Thanks @skedwards88 -- in that case I'll pin this to a patch version (1.10.0)

deployments/azure-webapps-node.yml
JasonFreeberg: Hi Sarah, users can download the deployment artifact from the Actions UI for debugging purposes. Also, it helps highlight the nice jobs UI and links.

image

deployments/azure-webapps-java-jar.yml
JasonFreeberg: @jeremyepling Good _cache_! 😉 

Do you want to enable caching for the other languages that don't support caching in the `setup-xxx` actions? On one hand it'll improve perf, but now sure if you want to expose all that in *starter* workflows.

```yaml
- name: Cache multiple paths
  uses: actions/cache@v2
  with:
    path: |
      ~/cache
      !~/cache/exclude
    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
 ```

File: deployments/azure-webapps-node.yml (hunk 5 above)
Comment (skedwards88) [reply deployments/azure-webapps-node.yml:52]: Awesome, thanks for sharing your reasoning!

File: deployments/azure-webapps-php.yml [hunk 7]
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          php-version: ${{ env.PHP_VERSION }}
+
+      - name: Check if composer.json exists
+        id: check_files
+        uses: andstor/file-existence-action@v1
+        with:
+          files: 'composer.json'
+
+      - name: Run composer install if composer.json exists
```
anuragc617, line 51: If composer.json won't exist then what artifacts will get published? I think then there won't be any artifacts. If that's the case should be remove the file exists condition

JasonFreeberg: If composer.json doesn't exist then the "raw" PHP files are deployed. This is here to determine if composer needs to be run for PHP apps which are using it. There's a separate class of PHP apps wqhich do not need composer, hence why we need the check

JasonFreeberg:
BTW @bishal-pdMSFT it says there is a merge conflict, but I can't fix it because it says that "Only those with write access to this repository can merge pull requests." Wanted to let you know because it looks like we're nearing the end of the reviews.

anuragc617 (approved):

deployments/azure-webapps-container.yml
```diff
@@ -0,0 +1,75 @@
...
+        with:
+          push: true
+          tags: ghcr.io/${{ env.REPO }}:${{ github.sha }}
+          file: ./Dockerfile
+
+    deploy:
+      runs-on: ubuntu-latest
+      needs: build
+      environment:
+        name: 'Development'
```
pulkitaggarwl, line 62: @JasonFreeberg  the environment should be "production" in this and other templates like it was before - https://github.com/actions/starter-workflows/blob/main/deployments/azure.yml#L32

ashwinsangem:
@JasonFreeberg, there's merge conflict for only one file now(azure.yml). This is because it was updated in the destination branch and deleted in your PR. Can you just pull the changes once from the destination branch and update your PR.

deployments/azure-webapps-dotnet-core.yml
```diff
@@ -0,0 +1,76 @@
+# This workflow will build and push a .NET Core app to an Azure Web App when a commit is pushed to your default branch.
+#
+# This workflow assumes you have already created the target Azure App Service web app.
+# For instructions see https://docs.microsoft.com/en-us/azure/app-service/quickstart-nodejs?tabs=linux&pivots=development-environment-cli
```
skedwards88, line 4: 🚗 Another drive by comment as I refer to this PR for the docs: You probably want to update the link here and elsewhere to the relevant guide.

deployments/azure-webapps-container.yml
JasonFreeberg: Atul gave the opposite advice: 

https://github.com/actions/starter-workflows/pull/1162#discussion_r733132170

https://github.com/actions/starter-workflows/pull/1162#discussion_r733132401

deployments/azure-webapps-dotnet-core.yml
JasonFreeberg: Good catch, thanks! Just fixed those links

deployments/azure-webapps-container.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Log in to GitHub container registry
+        uses: docker/login-action@v1.10.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
skedwards88, line 46: Can you also list the [permissions](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idpermissions) that are required for GITHUB_TOKEN for this job?

skedwards88:
@JasonFreeberg  You can find a preview of the corresponding guides for these starter workflows here: https://docs-internal-22832--ske-azure.herokuapp.com/en/actions/deployment/deploying-to-your-cloud-provider/deploying-to-azure The corresponding PR is https://github.com/github/docs-internal/pull/22832 (you will need to accept my invite to the docs-internal repo to view it). **I did not do any technical validation, so please leave any technical feedback.**

deployments/azure-webapps-container.yml
bishal-pdMSFT: @JasonFreeberg you can refer to this [doc](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow) to find what permission to specify

deployments/azure-webapps-java-jar.yml
bishal-pdMSFT: @JasonFreeberg starter workflows set a guiding example for user to create good workflows. Hence it would be okay to use the `actions/cache` action for such languages.
@pulkitaggarwl any thoughts?

pulkitaggarwl: @bishal-pdMSFT  this makes sense

anuragc617: @JasonFreeberg` actions/setup-<language>` support in-built caching eg. https://github.com/actions/setup-node#caching-packages-dependencies So, I think you can use the setup-<language> action for caching dependencies. And if setup action does support cache then you use `actions/cache` for caching.

JasonFreeberg: Just pushed a commit to add caching for .NET, Python, PHP, and Node workflows.

deployments/azure-webapps-container.yml
JasonFreeberg: @skedwards88 -- I was out sick, thanks for your patience. The required permissions are "repo" and "read:packages". I updated the workflow to use the default workflow token (`github.token`), but the PAT should still be created and provided to the Azure web app. I'll provide more details on the docs PR

skedwards88: @JasonFreeberg IIUC, `github.token` and `secrets.GITHU_TOKEN` are functionally equivalent, and permissions should be specified for both. I think you will want to add the `permissions` key to this workflow; otherwise, the workflow will fail for users who grant GITHUB_TOKEN limited scope by default.

skedwards88: @JasonFreeberg GITHUB_TOKEN permissions are different from PAT scopes. I am guessing that the permissions are ` contents: read` and `packages: write`, but confirmation before the corresponding guide for this template will be helpful: https://github.com/github/docs-internal/pull/22832

mehran-sayyah:
hi

PR #1199, opened 2021-10-27
Fix dotnet-desktop template

mattiascibien:
Removed environment variable which is not currently used and makes the build fail

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```


anuragc617 (approved):
Thanks @mattiascibien for fixing this. Changes looks good to me.

PR #1206, opened 2021-10-29
Deploy application to K8s Cluster

gambtho:
## Pre-requisites

- [X] Prior to submitting a new workflow, please apply to join the GitHub Technology Partner Program: [partner.github.com/apply](https://partner.github.com/apply?partnershipType=Technology+Partner).

---

---

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```

### DIFF
--- a/deployments/properties/k8s.properties.json
+++ b/deployments/properties/k8s.properties.json
@@ -1,0 +1,7 @@
+{
+    "name": "Deploy to a Kubernetes Cluster",
+    "description": "Deploy an application to a Kubernetes Cluster",
+    "creator": "Microsoft Azure",
+    "iconName": "kubernetes",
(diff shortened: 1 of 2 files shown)


deployments/properties/k8s.properties.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Deploy to a Kubernetes Cluster",
```
anuragc617, line 2: should we have this "Deploy to a AKS cluster using KubeConfig"

gambtho: no -- this is intentionally generic, as it works for any K8s cluster.  Adding "using kubeconfig" seems redundant, as that is always required to interact with the k8s api server

PR #1207, opened 2021-10-29
Deploy an application to AKS

gambtho:
## Tasks

**Some general notes:**

    ```
    # documentation.
    ```


deployments/properties/aks.properties.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Deploy to a AKS Cluster",
+    "description": "Deploy an application to a Azure Kubernetes Service Cluster",
+    "creator": "Microsoft Azure",
+    "iconName": "azure",
+    "categories": ["Deployment", "AKS", "Kompose", "Helm", "Kustomize", "Kubernetes"]
```
anuragc617, line 6: Kompose, Helm, kustomize are not currently detected by repo analysis. Here is the list of tech stacks that are detected https://github.com/github-starter-workflows/repo-analysis-partner/blob/main/tech_stacks.yml. You can raise the request to detect these here: https://github.com/github-starter-workflows/repo-analysis-partner/issues by providing how we should detect them.
Also, AKS is not a tech stack, I think we should remove it as it will not be used for suggesting the templates.

gambtho: ok, done
https://github.com/github-starter-workflows/repo-analysis-partner/issues/17
https://github.com/github-starter-workflows/repo-analysis-partner/issues/18
https://github.com/github-starter-workflows/repo-analysis-partner/issues/19

deployments/properties/aks.properties.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Deploy to a AKS Cluster",
+    "description": "Deploy an application to a Azure Kubernetes Service Cluster",
+    "creator": "Microsoft Azure",
+    "iconName": "azure",
+    "categories": ["Deployment", "Kompose", "Helm", "Kustomize", "Kubernetes"]
```
anuragc617, line 6: I noticed this later, this templates only support helm charts. Should we removed then Kompose and Kustomize as well from the categories.

anuragc617 (approved):

deployments/properties/aks.properties.json
gambtho: it supports all of them, but they have to make adjustments in the k8s-bake section as described in the readme for k8s-bake

gambtho: unsure if the starter workflow experience requires that to be automatic?   a bit tough as there are property name changes

anuragc617: Yes, then it means user will have to modify by themselves. In that case should we add a comment suggesting user to change properties accordingly.

anuragc617: Also, add  Dockerfile to the categories.

deployments/properties/aks.properties.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Deploy to a AKS Cluster",
```
anuragc617, line 2: Should it be named to "Deploy to a AKS Cluster using Azure Credentials", to distinguish it from template added in https://github.com/actions/starter-workflows/pull/1206

IUVSARINA2l1:
Shit how do you work this

deployments/properties/aks.properties.json
gambtho: the need to update the properties is already in the comments section at the top

deployments/properties/aks.properties.json
gambtho: Deploy to "AKS cluster" seems to distinguish it pretty clearly?

viccastro0916 (approved):

deployments/properties/aks.properties.json
```diff
@@ -0,0 +1,7 @@
...
+    "description": "Deploy an application to a Azure Kubernetes Service Cluster using Azure Credentials",
+    "creator": "Microsoft Azure",
+    "iconName": "azure",
+    "categories": ["Deployment", "Kompose", "Helm", "Kustomize", "Kubernetes, "Dockerfile"]
```
pulkitaggarwl, line 6: @gambtho 
1.  should we add Manifest to this list here ? 
2. The properties - Kompose, Helm, Kustomize, Kubernetes - are not being detected today. Can you update the detection logic to add those. And remove the properties from here which we cant detect ?  
3. the other template  - https://github.com/actions/starter-workflows/pull/1206/files  will always conflict with this one. I recommend to use only one of the templates. My understanding is that the two templates are differing wrt to authentication, if supporting both auth is really important, you might want to evaluate having one as default and other as comment. 
cc - @bishal-pdMSFT / @anuragc617

gambtho: - i added a PR to detect 3 of the 4...haven't seen any response to them
- happy to close the other template
- no idea how you'd want to detect manifest -- feel free to add detection logic, and i'll remove the other

- is it possible to get a consolidated PR review, rather than one-off comments every couple days?

gambtho: links to the detection PRs - https://github.com/actions/starter-workflows/pull/1207#discussion_r745551029
(in the comments above, from last week)

deployments/aks.yml
```diff
@@ -0,0 +1,79 @@
...
+      with:
+        creds: ${{ secrets.AZURE_CREDENTIALS }}
+    
+    - name: Build image on ACR
+      uses: azure/CLI@v1
+      with:
+        azcliversion: 2.29.1
+        inlineScript: |
+          az configure --defaults acr=${{ env.AZURE_CONTAINER_REGISTRY }}
+          az acr build -t  -t ${{ secrets.REGISTRY_URL }}/${{ env.PROJECT_NAME }}:${{ github.sha }}
```
skedwards88, line 51: 👋 I'm writing the corresponding guide for this template. Can you tell me why `REGISTRY_URL` is a secret instead of an env variable (like `PROJECT_NAME`)?

skedwards88:
@gambtho You can find a preview of the corresponding guide for this starter workflow here: https://docs-internal-22832--ske-azure.herokuapp.com/en/actions/deployment/deploying-to-your-cloud-provider/deploying-to-azure/deploying-to-azure-kubernetes-service The corresponding PR is https://github.com/github/docs-internal/pull/22832 (you will need to accept my invite to the docs-internal repo to view it). **I did not do any technical validation, so please leave any technical feedback.**

deployments/aks.yml
gambtho: no good reason -- could move it to env variable (it's an artifact from a different workflow where there was a registry secret, but not valid here)

deployments/properties/aks.properties.json
gambtho: @pulkitaggarwl what's the best way to have one as a comment?   have it hosted in a separate repo, and link to it?

pulkitaggarwl: I would suggest adding some comments and routing them to documentation pages with details around second one

gambtho: sorry -- i'm not following.   Adding comments to the properties file, and routing them to where?

Tanapl02:
> ## Tasks
> 
> **For _all_ workflows, the workflow:**
> 
> - [x] Should be contained in a `.yml` file with the language or platform as its filename, in lower, [_kebab-cased_](https://en.wikipedia.org/wiki/Kebab_case) format (for example, [`docker-image.yml`](https://github.com/actions/starter-workflows/blob/main/ci/docker-image.yml)).  Special characters should be removed or replaced with words as appropriate (for example, "dotnet" instead of ".NET").
> - [x] Should use sentence case for the names of workflows and steps (for example, "Run tests").
> - [x] Should be named _only_ by the name of the language or platform (for example, "Go", not "Go CI" or "Go Build").
> - [x] Should include comments in the workflow for any parts that are not obvious or could use clarification.
> 
> 
> **Some general notes:**
> 
> - [x] This workflow must _only_ use actions that are produced by GitHub, [in the `actions` organization](https://github.com/actions), **or**
> - [x] This workflow must _only_ use actions that are produced by the language or ecosystem that the workflow supports.  These actions must be [published to the GitHub Marketplace](https://github.com/marketplace?type=actions).  We require that these actions be referenced using the full 40 character hash of the action's commit instead of a tag.  Additionally, workflows must include the following comment at the top of the workflow file:
>     ```
>     # This workflow uses actions that are not certified by GitHub.
>     # They are provided by a third-party and are governed by
>     # separate terms of service, privacy policy, and support
>     # documentation.
>     ```
> - [x] Automation and CI workflows should not send data to any 3rd party service except for the purposes of installing dependencies.
> - [x] Automation and CI workflows cannot be dependent on a paid service or product.
> 

[]()start

deployments/aks.yml
gambtho: removed this variable as uneeded - since we had the acr registry name already

deployments/properties/aks.properties.json
pulkitaggarwl: @gambtho, sure I can elaborate. 
My suggestion was to add some comments in the 1st YML file highlighting that there is an alternate  CI / CD to AKS can be used.  An azure docs page with the alternate version and details of the YML components of the alternate version can be used.

pulkitaggarwl (approved):

deployments/properties/aks.properties.json
gambtho: ahah, makes sense to me.  i'll close the other one for now, and will submit a new PR once we have a docs home for it

PR #1255, opened 2021-11-17
Have the starter `docker-publish` action sign digests.

mattmoor:
This change installs `sigstore/cosign` using the `cosign-installer` action,
and uses sigstore's "keyless" signing process to sign the resulting image
digest using the action's identity token (see: `id-token: write`).

Signed-off-by: Matt Moore <mattomata@gmail.com>

## Pre-requisites

---

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```

### DIFF
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -43,6 +43,10 @@ jobs:
         if: github.event_name != 'pull_request'
         uses: sigstore/cosign-installer@1e95c1de343b5b0c23352d6417ee3e48d5bcd422
 
+      # Workaround: https://github.com/docker/build-push-action/issues/461
+      - name: Setup Docker buildx
+        uses: docker/setup-buildx-action@79abd3f86f79a9d68a23c75a09a9a85889262adf
+
       # Login against a Docker registry except on PR
       # https://github.com/docker/login-action
       - name: Log into registry ${{ env.REGISTRY }}
@@ -81,4 +85,4 @@ jobs:
         # This step uses the identity token to provision an ephemeral certificate
         # against the sigstore community Fulcio instance, and records it to the
         # sigstore community Rekor transparency log.
-        run: cosign sign --force ${{ steps.build-and-push.outputs.digest }}
+        run: cosign sign --force ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}@${{ steps.build-and-push.outputs.digest }}
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -76,13 +76,27 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
 
-      # Sign the resulting Docker image digest except on PR
+      # Check whether this is a public repository before signing the image.
+      # The keyless signing process records signatures on the Rekor public
+      # transparency log, so signing is disabled for private repos by default
+      # to avoid leaking private data.  If you wish to sign things anyways,
+      # then this check can be removed and --force can be added to the cosign
+      # command below.
+      - name: Confirm this is a public repository
+        if: ${{ github.event_name != 'pull_request' }}
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        run: |
+          TEMP=$(gh api repos/$GITHUB_REPOSITORY --jq '.private')
+          echo "IS_PRIVATE=$TEMP" >> $GITHUB_ENV
+
+      # Sign the resulting Docker image digest except on PRs and private repos
       # https://github.com/sigstore/cosign
       - name: Sign the published Docker image
-        if: github.event_name != 'pull_request'
+        if: ${{ github.event_name != 'pull_request' && env.IS_PRIVATE == 'false' }}
         env:
           COSIGN_EXPERIMENTAL: "true"
         # This step uses the identity token to provision an ephemeral certificate
         # against the sigstore community Fulcio instance, and records it to the
         # sigstore community Rekor transparency log.
-        run: cosign sign --force ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}@${{ steps.build-and-push.outputs.digest }}
+        run: cosign sign ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}@${{ steps.build-and-push.outputs.digest }}
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -76,24 +76,15 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
 
-      # Check whether this is a public repository before signing the image.
+      # Sign the resulting Docker image digest except on PRs and private repos
       # The keyless signing process records signatures on the Rekor public
       # transparency log, so signing is disabled for private repos by default
       # to avoid leaking private data.  If you wish to sign things anyways,
       # then this check can be removed and --force can be added to the cosign
       # command below.
-      - name: Confirm this is a public repository
-        if: ${{ github.event_name != 'pull_request' }}
-        env:
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-        run: |
-          TEMP=$(gh api repos/$GITHUB_REPOSITORY --jq '.private')
-          echo "IS_PRIVATE=$TEMP" >> $GITHUB_ENV
-
-      # Sign the resulting Docker image digest except on PRs and private repos
       # https://github.com/sigstore/cosign
       - name: Sign the published Docker image
-        if: ${{ github.event_name != 'pull_request' && env.IS_PRIVATE == 'false' }}
+        if: ${{ github.event_name != 'pull_request' && !github.event.repository.private }}
         env:
           COSIGN_EXPERIMENTAL: "true"
         # This step uses the identity token to provision an ephemeral certificate
--- a/ci/docker-publish.yml
+++ b/ci/docker-publish.yml
@@ -29,10 +29,19 @@ jobs:
     permissions:
       contents: read
       packages: write
+      # This is used to complete the identity challenge
+      # with sigstore/fulcio when running outside of PRs.
+      id-token: write
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+
+      # Install the cosign tool except on PR
+      # https://github.com/sigstore/cosign-installer
+      - name: Install cosign
+        if: github.event_name != 'pull_request'
+        uses: sigstore/cosign-installer@1e95c1de343b5b0c23352d6417ee3e48d5bcd422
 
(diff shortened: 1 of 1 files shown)


ci/docker-publish.yml
```diff
@@ -29,11 +29,20 @@ jobs:
...
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
 
+      # Install the cosign tool except on PR
+      # https://github.com/sigstore/cosign-installer
+      - name: Install cosign
+        if: github.event_name != 'pull_request'
+        uses: sigstore/cosign-installer@1e95c1de343b5b0c23352d6417ee3e48d5bcd422
```
dekkagaijin, line 44:
```suggestion
        uses: sigstore/cosign-installer@main
```

To be consistent with our guidance? Unless the style for these examples requires pinning to a git commit

mattmoor: I was trying to be consistent with the other steps, which do this.

jhutchings1:
@andymckay @chrispat This workflow sends data to a 3rd party service for the purposes of providing a public log of the signing activity. I think this is a good reason to do it, but I'd love your insight on how best to set expectations on this for customers.

ci/docker-publish.yml
chrispat: We require that starter-workflows pin to a specific commit when including actions not maintained by GitHub.  In addition pinning to a ref like main is not recommended as updates could break customers.

dekkagaijin: @chrispat since we are a security-focused action, we recommend using `@main`for the action to adopt fixes ASAP
and to optionally pin a specific release version of the tool itself if that degree of stability is desired

mattmoor: As folks involved in supply chain security, we also understand a desire to be very intentional about picking up changes to dependencies 😉

mattmoor: Maybe encoding the `@main` recommendation in a comment above that step is a reasonable compromise?

ethomson: You can see what Ruby does: https://github.com/actions/starter-workflows/blob/main/ci/ruby.yml#L29

(I'd encourage you to encourage pinning to a released tag, not `@main`, that's more Actions-idiomatic.)

ci/docker-publish.yml
```diff
@@ -55,9 +68,21 @@ jobs:
...
         with:
           context: .
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+      # Sign the resulting Docker image digest except on PR
+      # https://github.com/sigstore/cosign
+      - name: Sign the published Docker image
+        if: github.event_name != 'pull_request'
```
jhutchings1, line 82: Was chatting with @steiza, and he brought up an interesting point here. For private repositories, the transparency log will effectively leak the identity of a private repository. We had the thought that we should add an if to only sign for public repositories. Users could obviously opt-in to signing for private repositories, but this would not leak their private repo information by default. 

@mattmoor what do you think?

mattmoor: That works for me if there's a way to express that in actions (I think the best UX here would be for it to show up as skipped like it would for PRs).

I suspect the next best would be for us to add an option to cosign that lets us toggle this off (perhaps with a clear warning) when the image is private (I can discuss with @dlorenc @dekkagaijin et al and try to land something in 1.4).

WDYT?

mattmoor: I staged a fix in cosign for this here: https://github.com/sigstore/cosign/pull/1116, which would let us drop the `--force` (although I'll probably add a comment indicating that folks comfortable with this can add `--force` to publish to the transparency log).

jhutchings1: Yep, it's very doable. You could add the step below, and then add the (!(env.IS_PRIVATE)) to your existing if statement and I think it would do the job. 

```yaml

      - name: Confirm this is a public repository
        env: 
          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
        run: |
          TEMP=$(gh api repos/$GITHUB_REPOSITORY --jq '.private')
          echo "IS_PRIVATE=$TEMP" >> $GITHUB_ENV
      - name: Run if public
        if: (!(env.IS_PRIVATE))

```

mattmoor: Awesome TIL!  I'll try that out and then update this.  I realize that this'll end up skipping signing entirely, so we may still want to adopt the cosign change and drop this once we cut 1.4.

mattmoor: ok, I was able to get this working in my little sample repo:
```yaml
      - name: Confirm this is a public repository
        env:
          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
        run: |
          TEMP=$(gh api repos/$GITHUB_REPOSITORY --jq '.private')
          echo "IS_PRIVATE=$TEMP" >> $GITHUB_ENV

      - name: Sign the container image (if public)
        if: ${{ env.IS_PRIVATE == 'false' }}
        env:
          COSIGN_EXPERIMENTAL: "true"
        run: cosign sign ghcr.io/${{ github.repository }}@${{ steps.push-step.outputs.digest }}
```

This ran fine (with signing) in the public repo, and from a private repo instantiated from it the signing step was skipped.  I'll update the PR with this shortly, thanks for the tip!

mattmoor: Ok, cool this should be updated.  When we cut cosign `1.4` we can simplify things, since my PR upstream landed.

chrispat: Given you have the whole event payload to work with already on disk you don't need to make the api call.  Each event has a repository object and that will tell you if it is public or private.

ci/docker-publish.yml
```diff
@@ -55,9 +68,35 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        run: |
+          TEMP=$(gh api repos/$GITHUB_REPOSITORY --jq '.private')
+          echo "IS_PRIVATE=$TEMP" >> $GITHUB_ENV
+
+      # Sign the resulting Docker image digest except on PRs and private repos
+      # https://github.com/sigstore/cosign
+      - name: Sign the published Docker image
+        if: ${{ github.event_name != 'pull_request' && env.IS_PRIVATE == 'false' }}
```
steiza, line 96: I came here to suggest that we only run this on public repos, but you beat me to it!

I think we could simplify the implementation a bit though, by not having the above step and just checking `if: ${{ !github.event.repository.private }}` here.

It shouldn't be the default, but there might be use-cases where a private repo wants to use keyless signing, and is okay with the public record, so we could have an opt-in env variable (`COSIGN_USE_PUBLIC_KEYLESS_SIGNING`?) that uses keyless signing regardless. Not required, just a suggestion.

Thanks for contributing this functionality!

mattmoor: @steiza even better, I'll update the PR.  The next version of `cosign` will make it so we can always sign and for private repos they would add `--force` if they want to publish to the transparency log, which feels like exactly the opt-in we want.

mattmoor: Perfect, this is much tighter, and I just tried it on my public and private demo repos and it works like a charm.  ✨ 

Thanks again @steiza !

ci/docker-publish.yml
mattmoor: Looks like what @steiza also said.  This should be fixed in the latest revision!

jhutchings1:
Quick update for everyone: I'm working on a communications plan so that we notify users when this change goes in. I'd like to target Monday 12/6 to get everything out, but we might need to move things around depending on internal reviews. Really appreciate all the collaboration here!

mattmoor:
Cool, I'll try to keep things rebased 👍

jhutchings1 (approved):
Changes look good to me. Thanks for the collaboration all!

bishal-pdMSFT (approved):

mehran-sayyah:
#1255

دریافت Outlook برای Android<https://aka.ms/AAb9ysg>
________________________________

bishal-pdMSFT:
@mattmoor the updated workflow failed for my docker [repo](https://github.com/bishal-pdMSFT/docker-app/runs/4434089694?check_suite_focus=true). 
If failed in signing step:
```
Error: signing [ghcr.io/bishal-pdMSFT/docker-app@sha256:ec9994d1d8cf472600bbc9fa73d2e6afe865beef000185659f6d0e8f774b47f8]: parsing reference: could not parse reference: ghcr.io/bishal-pdMSFT/docker-app@sha256:ec9994d1d8cf472600bbc9fa73d2e6afe865beef000185659f6d0e8f774b47f8
main.go:46: error during command execution: signing [ghcr.io/bishal-pdMSFT/docker-app@sha256:ec9994d1d8cf472600bbc9fa73d2e6afe865beef000185659f6d0e8f774b47f8]: parsing reference: could not parse reference: ghcr.io/bishal-pdMSFT/docker-app@sha256:ec9994d1d8cf472600bbc9fa73d2e6afe865beef000185659f6d0e8f774b47f8
```

Can you please check it at priority and let me know if this can be widespread?

mattmoor:
I will check it out now, I was in a meeting when things merged, but I wonder if it is the capitals in your username, which I don't believe are allowed in Docker repository names (not sure how ghcr.io maps this?)

mattmoor:
Looks like that's the case:
```shell
# crane ls ghcr.io/bishal-pdMSFT/docker-app
2021/12/06 10:15:51 reading tags for ghcr.io/bishal-pdMSFT/docker-app: parsing repo "ghcr.io/bishal-pdMSFT/docker-app": repository can only contain the runes `abcdefghijklmnopqrstuvwxyz0123456789_-./`: bishal-pdMSFT/docker-app
```

Depending on how this is handled by `ghcr.io` we should be able to get this fixed pretty quickly 🤞

mattmoor:
Looks like buildkit is implicitly making the username lowercase: https://github.com/bishal-pdMSFT/docker-app/runs/4434089694?check_suite_focus=true#step:7:187

I assume this is roughly what we want here.  Are github usernames guaranteed to be unique regardless of case?

mattmoor:
I confirmed that this works for me here: https://github.com/mattmoor/zero-friction-actions/runs/4434363881?check_suite_focus=true

Let me open an issue here to track the uppercase issue, and discuss options for fixes.

bishal-pdMSFT:
Looks like account name is case insensitive. I tried changing my account name to another name with different case and it was not allowed.

bishal-pdMSFT:
Confirmed internally as well. It is case insensitive.

bishal-pdMSFT:
And for the first time I realize the benefit of choosing a CAPS letter in my account name :grin:

mattmoor:
Looks like this is `docker/metadata-action` lowercasing things, so I posted https://github.com/actions/starter-workflows/pull/1294 which should fix this.  Sorry for the breakage @bishal-pdMSFT, and thanks for the pointer @imjasonh

PR #1281, opened 2021-12-01
chore: split npm publish into 2 workflows

MylesBorins:
Currently we suggest that folks dual publish to both npm + gpr.

There are a large number of edge cases related to doing this and IMHO it is
not the best practice. Let's make two separate workflows.

### DIFF
--- a/ci/npm-publish-github-packages.yml
+++ b/ci/npm-publish-github-packages.yml
@@ -18,16 +18,19 @@ jobs:
       - run: npm ci
       - run: npm test
 
-  publish-npm:
+  publish-gpr:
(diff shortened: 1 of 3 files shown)


ethomson (approved):

joshmgross (approved):

bishal-pdMSFT:
@MylesBorins looking at a repo is there a way to differentiate which template should be preferred?

MylesBorins:
@bishal-pdMSFT customers should pick the template based on the registry they want to publish to. The majority of customers will want to publish to the public registry, not ghp

Issue #1299, opened 2021-12-08
BUG: starter-workflows does not ask workflows to declare permissions


laurentsimon:
GitHub asks users to define workflow permissions, see https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/ and https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

Until recently, this repo did not follow the recommendations https://github.com/actions/starter-workflows/pull/1072/files. Thanks @varunsh-coder  for helping out!

This repo is home to many "partnered" GitHub actions. However, it does not enforce the workflows follow these best practices.

I ran OSSF [scorecard](github.com/ossf/scorecard) and found that most of the 80+ starter workflows don't declare their permissions. Just one example for ref https://github.com/actions/starter-workflows/blob/main/code-scanning/checkmarx.yml

Since this affects all users of the workflows, I think permissions should be enforced during PR reviews and maybe thru tooling to verify the permissions adhere to the principle of least privilege.

laurentsimon:
cc @josepalafox

josepalafox:
Nothing quite like learning you're the emperor with no clothes. Thank you for this, we'll have some internal discussions here. @pulkitaggarwl

varunsh-coder:
Thanks @laurentsimon for tagging me! JFYI - I am working on a way for developers to set token permissions automatically in their workflows. The solution is [open source](https://github.com/step-security/secure-workflows), can be used from a [website](https://app.stepsecurity.io) (no installation needed), and is based on a [knowledge-base (KB)](https://github.com/step-security/secure-workflows/tree/main/knowledge-base) of different GitHub Actions. Would be great if you can suggest developers to use it to set token permissions (as part of the process that is setup to resolve this issue). Would be easier for them to do than to calculate permissions manually. It will also help improve the KB...in case the submitter owns a GitHub Action, they can add to the KB, which will benefit others...

varunsh-coder:
I can fix token permissions for all the workflows while building out the KB (mentioned in comment above). would you prefer changes to be done for each workflow in separate PR? or should I fix multiple workflows in a single PR?

bishal-pdMSFT:
@varunsh-coder thanks for pointing out the gap in starter workflows! 🙇 
I will add a point in the PR template to call this out. And please feel free to raise PR for other workflows to fix their permission scope.

varunsh-coder:
[...]

PR #1341, opened 2022-01-05
Add token permissions for ci/erlang.yml

varunsh-coder:
This is related to issue #1299. This PR adds token permissions for  ci/erlang.yml. 

## Pre-requisites

---

---

## Tasks

- [ ] Should be contained in a `.yml` file with the language or platform as its filename, in lower, [_kebab-cased_](https://en.wikipedia.org/wiki/Kebab_case) format (for example, [`docker-image.yml`](https://github.com/actions/starter-workflows/blob/main/ci/docker-image.yml)).  Special characters should be removed or replaced with words as appropriate (for example, "dotnet" instead of ".NET").
- [ ] Should use sentence case for the names of workflows and steps (for example, "Run tests").
- [ ] Should be named _only_ by the name of the language or platform (for example, "Go", not "Go CI" or "Go Build").
- [ ] Should include comments in the workflow for any parts that are not obvious or could use clarification.

- [ ] Should be preserved under [the `code-scanning` directory](https://github.com/actions/starter-workflows/tree/main/ci).
- [ ] Should include a matching `code-scanning/properties/*.properties.json` file (for example, [`code-scanning/properties/codeql.properties.json`](https://github.com/actions/starter-workflows/blob/main/code-scanning/properties/codeql.properties.json)), with properties set as follows:
  - [ ] `name`: Name of the Code Scanning integration.
  - [ ] `organization`: Name of the organization producing the Code Scanning integration.
  - [ ] `description`: Short description of the Code Scanning integration.
  - [ ] `categories`: Array of languages supported by the Code Scanning integration.
  - [ ] `iconName`: Name of the SVG logo representing the Code Scanning integration. This SVG logo must be present in [the `icons` directory](https://github.com/actions/starter-workflows/tree/main/icons).
- [ ] Should run on `push` to `branches: [ $default-branch, $protected-branches ]` and `pull_request` to `branches: [ $default-branch ]`. We also recommend a `schedule` trigger of `cron: $cron-weekly` (for example, [`codeql.yml`](https://github.com/actions/starter-workflows/blob/c59b62dee0eae1f9f368b7011cf05c2fc42cf084/code-scanning/codeql.yml#L14-L21)).

**Some general notes:**

    ```
    # documentation.
    ```


ci/erlang.yml
```diff
@@ -6,10 +6,14 @@ on:
   pull_request:
     branches: [ $default-branch ]
 
+permissions: read-all
```
bishal-pdMSFT, line 9: Why is this needed?

ci/erlang.yml
```diff
@@ -6,10 +6,14 @@ on:
...
+
 jobs:
 
   build:
 
+    permissions:
+      contents: read # for actions/checkout to fetch code
```
bishal-pdMSFT, line 16: I believe `contents: read` is a default [permission](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) even for restricted access.
So in a strict sense this change is not needed, and that keeps the workflow simple.

ci/erlang.yml
varunsh-coder: This permission at the workflow (top) level is so that any future jobs that get added to the same workflow are secure by default. It also makes the changes in-line with expectation from [ossf/scorecard](https://github.com/ossf/scorecard). You can see the discussion [here](https://github.com/ossf/scorecard/issues/1128#issuecomment-973236716). 

CC: @laurentsimon

ci/erlang.yml
varunsh-coder: My understanding is that when a new GitHub org or repo is created, the default GitHub Action token permissions are permissive. By that I mean that all permissions are granted to the `GITHUB_TOKEN`. So, if `contents: read` is not added in the workflow file, the token will get all permissions. 

Moreover, while writing the starter workflows, we do not know what is the token permission set at the org/ repo level for whoever will use the workflow. So, the permissions should be specified in the workflow file.

bishal-pdMSFT: > So, if contents: read is not added in the workflow file, the token will get all permissions.

That's a good point in case of permissive repos/orgs.

ci/erlang.yml
bishal-pdMSFT: The discussion ⬆️ was good read. Thanks!
And I agree it makes sense to add permission at workflow level for new jobs.

ci/erlang.yml
bishal-pdMSFT: However, given that we have already specified a workflow level read permission, this job level permission seems redundant.
I am not against good practice of explicit permissions, but this one here is just adding one more line which is not much useful. Simplicity of workflow is important, especially in this repo as it is meant for the users to get started with workflows.

varunsh-coder: @bishal-pdMSFT this is good feedback. Since the changes are done via automation, I would want to address this via a configuration option in the automation. What if there is a configuration option to `Only set job level permissions`. In this case, the workflow level (top level) permissions will not be set, and only job level permissions will be set. So the workflow would be simpler. Moreover, the workflow would get a score of 9 in scorecards. Not perfect from a security perspective, but it balances security with simplicity. Will this work?

CC: @laurentsimon

bishal-pdMSFT: Can we add smartness in automation to `Set workflow level permissions` and `Set job level permissions if they are not same as default`?

varunsh-coder: @bishal-pdMSFT did you mean `Set job level permissions if they are not same as workflow level permissions`? As of now, the automation sets the workflow level permissions to `read-all`. So, in this case, if the job level permission is not set, it would be `read-all`, which is more than `contents: read` (current value). `read-all` will include all read permissions, including `SecurityEvents` - example [here](https://github.com/varunsh-coder/actions-playground/runs/4844452755?check_suite_focus=true#step:1:13).

bishal-pdMSFT: I guess that should be fine atleast for a GitHub support action like `actions/checkout`. A job containing a third party action may decide to restrict using `contents: read`.
Again it is a trade off between security and simplicity.

varunsh-coder: @bishal-pdMSFT to clarify, are you suggesting only having the `permissions: read-all` permission at workflow level, and removing the permissions from job level? If that is the case, the configuration could be something like `Only add job level permissions if they are higher than read-all`.

laurentsimon: if there's no write permission needed for the job, the `read-all` top-level permission does simplify the workflow, I agree. You'd still get scorecard 10, correct?

varunsh-coder: @laurentsimon please see feedback on [this PR](https://github.com/Automattic/vip-go-mu-plugins/pull/2815#discussion_r786933173). The ask is to set `contents: read` at the workflow (top) level instead of `permissions: read-all`. This is what the repo level restrictive setting is (screenshot [here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions)), so it would align with that setting. Then I can update the StepSecurity automation such that if job level permission is `contents: read`, then do not add it explicitly (inherit from workflow). I believe this is aligned with what @bishal-pdMSFT is also asking for. What are your thoughts? I don't think you need to make changes to scorecards for this.

laurentsimon: `contents: read` sgtm.  If additionally read permissions are needed for a job within the workflow, you'll add them all at the job level, correct?

varunsh-coder: > `contents: read` sgtm. If additionally read permissions are needed for a job within the workflow, you'll add them all at the job level, correct?

Yes. Thanks! I will make this change.

varunsh-coder: @bishal-pdMSFT the logic has been updated to add `contents: read` at workflow level, and if job needs `contents: read` only, then to set no permissions at the job level. Applied changes to this yaml file. Let me know if the update looks fine.

ci/erlang.yml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ $default-branch ]
 
+permissions:
```
anuragc617, line 9: It's the default permission. Do we need it?

ci/erlang.yml
varunsh-coder: @bishal-pdMSFT - reminder to please review this. Thanks!

anuragc617 (approved):
Looks good to me.

PR #1353, opened 2022-01-14
Upgrade Rails workflow to true CI

indirect:
The existing Rails CI example only runs linters, which is not continuous integration. This change brings the Rails example workflow up to par with the other web framework CI flows, like Django.

This example is optimized for Rails 7, which does not include NodeJS, webpack, or yarn by default. No Rails application code changes are required for this flow to run the tests, and both minitest and rspec are supported via the `test` rake task.


ci/rubyonrails.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+      # Add or replace dependency steps here
+      - name: Install Ruby and gems
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      # Add or replace database setup steps here
+      - name: Set up database schema
+        run: bin/rails db:schema:load
```
bishal-pdMSFT, line 35: Should we also do `db:seed`?

indirect: The `db:seed` task is generally not for generating test data. It is for setting up local development databases with sample data. See [this discussion with the Rails core team](https://discuss.rubyonrails.org/t/should-dbprepare-also-call-db-seed-by-default/74835) for more information.

ci/rubyonrails.yml
```diff
@@ -0,0 +1,61 @@
...
+    services:
+      postgres:
+        image: postgres:11-alpine
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_DB: rails_test
+          POSTGRES_USER: rails
+          POSTGRES_PASSWORD: password
+    steps:
```
bishal-pdMSFT, line 25: Specify these values as [environment variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables) and use that across steps.

ci/rubyonrails.yml
```diff
@@ -0,0 +1,61 @@
...
+  pull_request:
+    branches: [ $default-branch ]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11-alpine
+        ports:
+          - 5432:5432
```
bishal-pdMSFT, line 20: Port can be environment variable as well

indirect: Why should this be an environment variable? This is the default post used by Postgres on every machine in existence. Also, this port number is only used in one place, the DATABASE_URL below.

ci/properties/rubyonrails.properties.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Ruby on Rails continuous integration",
+    "description": "Build, lint, and test a Rails application",
+    "iconName": "rails",
```
bishal-pdMSFT, line 4: This icon needs to be added to https://github.com/actions/starter-workflows/tree/main/icons directory.

ci/rubyonrails.yml
indirect: Why should these be environment variables? The entire point is that they should not be configured. The database is configured with the exact values that are hardcoded in the test runner step.

If someone wants to configure their workflow to use different values, that is fine, they can do that! This is a template. But the point is that this action should work immediately, without making any changes to the user's system, and without making any changes to the Rails application code or configuration.

ci/properties/rubyonrails.properties.json
indirect: Done, my apologies. It seems I had a bad gitignore line that prevented me from adding the icon previously.

indirect:
Okay, I've updated it to use environment variables. Just to warn you, I am sure it worked correctly before, because I tested it with my own Rails app. With the environment variable changes, I am no longer completely sure everything will work correctly. Hopefully it will, though!

ci/rubyonrails.yml
```diff
@@ -0,0 +1,66 @@
...
+        env:
+          POSTGRES_DB: ${{env.DB}}
+          POSTGRES_USER: ${{env.USER}}
+          POSTGRES_PASSWORD: ${{env.PASSWORD}}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      # Add or replace dependency steps here
+      - name: Install Ruby and gems
+        uses: ruby/setup-ruby@v1
```
bishal-pdMSFT, line 35: Please use the full SHA for any action which is not directly published by GitHub. See "general notes" section [here](https://github.com/actions/starter-workflows/blob/main/.github/pull_request_template.md).

indirect: Done.

indirect:
@bishal-pdMSFT it seems environment variables cannot be used for `services` as you requested. I tried it, and the workflow file is invalid: https://github.com/indirect/feedyour.email/actions/runs/1703936322/workflow#L21. There is also a forum discussion that seems to imply environment variables can only be used inside steps, which means it is impossible to share between the service and the step: https://github.community/t/how-to-use-env-context/16975/5.

indirect:
@bishal-pdMSFT after testing the latest version that removes ENV duplication, it seems it is also impossible to refer to job env vars inside the job env var section. I have tested the workflow and resolved the error, but it seems it is not possible to arrange job steps and services to share data via environment variables. Here is the successful workflow run without the env var error, with a workflow matching the latest commit in this PR: https://github.com/indirect/feedyour.email/actions/runs/1703962976

bishal-pdMSFT:
Thanks for pointing this out! The new changes look good to me.

bishal-pdMSFT:
Though I will wait for either @aparna-ravindra or @anuragc617 to take a look before we merge.

ci/rubyonrails.yml
```diff
@@ -0,0 +1,58 @@
...
+          POSTGRES_PASSWORD: password
+    env:
+      RAILS_ENV: test
+      DATABASE_URL: "postgres://rails:password@localhost:5432/rails_test"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      # Add or replace dependency steps here
+      - name: Install Ruby and gems
+        uses: ruby/setup-ruby@180e1d1f1c5eadbfbab3dab6fb79ab0c07e3cecc
```
bishal-pdMSFT, line 33: I do not see this commit SHA in any of the [releases](https://github.com/ruby/setup-ruby/releases) for this action. It would be good use a commit which has been officially released.

indirect: Ok, updated to the SHA of v1.92.

ci/rubyonrails.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Run tests
+        run: bin/rake
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install Ruby and gems
+        uses: ruby/setup-ruby@v1
```
aparna-ravindra, line 49: The same commit id used in the test job could be used here as well

ci/rubyonrails.yml
```diff
@@ -0,0 +1,58 @@
...
+  push:
+    branches: [ $default-branch ]
+  pull_request:
+    branches: [ $default-branch ]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11-alpine
```
aparna-ravindra, line 18: Do we need a health check to ensure that the service has started before it is used?

indirect: I don't think so: I have run this workflow about 250 times now, and the alpine image starts up so fast I have never seen an error.

ci/rubyonrails.yml
indirect: Ah, right. I will add it here as well.

aparna-ravindra:
LGTM

bishal-pdMSFT (approved):

bishal-pdMSFT:
@indirect let us know if you are fine for this to be merged?

indirect:
@bishal-pdMSFT looks good, let’s do it 👍🏻

bishal-pdMSFT:
@indirect merged. A huge thanks ❤️  for making starter workflows better 🎉

PR #1369, opened 2022-01-27
Fix version of `denoland/setup-deno`

arnu515:
The version of the `denoland/setup-deno` action is a wierd hash that doesn't work. Instead using `v1` from the marketplace listing seems to work.

### DIFF
--- a/ci/deno.yml
+++ b/ci/deno.yml
@@ -23,8 +23,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup Deno
-        # uses: denoland/setup-deno@v1
-        uses: denoland/setup-deno@004814556e37c54a2f6e31384c9e18e9833173669
+        uses: denoland/setup-deno@v1
         with:
           deno-version: v1.x
 
--- a/ci/deno.yml
+++ b/ci/deno.yml
@@ -23,7 +23,8 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup Deno
-        uses: denoland/setup-deno@v1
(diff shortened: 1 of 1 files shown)


ci/deno.yml
```diff
@@ -23,8 +23,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup Deno
-        # uses: denoland/setup-deno@v1
-        uses: denoland/setup-deno@004814556e37c54a2f6e31384c9e18e9833173669
+        uses: denoland/setup-deno@v1
```
joshmgross, line 26: > The version of the `denoland/setup-deno` action is a wierd hash that doesn't work. Instead using `v1` from the marketplace listing seems to work.

We're intentionally using a commit SHA for all 3rd party actions - see https://github.com/actions/starter-workflows/blob/main/CONTRIBUTING.md

That SHA was intended to be the commit SHA for https://github.com/denoland/setup-deno/releases/tag/v1.1.0
https://github.com/denoland/setup-deno/commit/004814556e37c54a2f6e31384c9e18e983317366

https://github.com/actions/starter-workflows/pull/1024/files#r676723335
```suggestion
        # uses: denoland/setup-deno@v1
        uses: denoland/setup-deno@004814556e37c54a2f6e31384c9e18e983317366
```

It looks like a typo added an extra digit to the end

joshmgross (changes requested):

arnu515:
It was erroring out for me, but it shouldn't anymore. Thanks.

joshmgross (approved):
Thanks for your contribution 🙇

PR #1402, opened 2022-02-11
Fix some workflows not being excluded from Enterprise syncing.

chrisgavin:
This allows manually excluding starter workflows from being bundled into an Enterprise Server release and removes a workflow that was mistakenly.

A lot of starter workflows won't work on Enterprise Server without specific tweaks to ensure it will work on self-hosted runners and if the runner is air-gapped. These starter workflows are currently detected based on the Actions they use, but in some cases there are workflows that don't include any Actions but use shell scripts to make outbound calls that might fail in an Enterprise Server environment.

In these cases it's useful to be able to manually specify that a workflow shouldn't be included in Enterprise Server. This adds the ability to do that.

### DIFF
--- a/script/sync-ghes/index.ts
+++ b/script/sync-ghes/index.ts
@@ -21,6 +21,8 @@ interface WorkflowProperties {
   categories: string[] | null;
 
   creator?: string;
+
+  enterprise?: boolean;
 }
 
 interface WorkflowsCheckResult {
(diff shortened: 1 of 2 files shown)


script/sync-ghes/index.ts
```diff
@@ -59,6 +61,7 @@ async function checkWorkflows(
 
         const enabled =
           !isPartnerWorkflow &&
+          workflowProperties.enterprise !== false &&
```
joshmgross, line 64:
```suggestion
          workflowProperties.enterprise &&
```

Do we need the explicit `!== false`?

chrisgavin: I think we do, yes. The field can also be `undefined` which I think we want to treat the same as it being true.

nickfyson (approved):
🚀

PR #1500, opened 2022-03-24
Add Security Gallery entry for new Dependency Review Action

elireisman:
As stated in the title. I cut this PR locally before having access to the `starter-workflows` repo so I didn't see this PR template and instructions until now. I will review and make updates as needed  before taking this out of draft, but feel free to call out anything I missed!

The goal is to prep a PR for merge that adds the new `actions/dependency-review-action` repo to the Security Gallery tab. We plan to create that repo by EOW and move our staff-shipped code from [here](https://github.com/dsp-testing/dependency-review-action) to there after a sanitation pass to prep the Git history (etc) for GA and open sourcing 👍 

## Pre-requisites

## Tasks

- [x] Should specify least priviledge [permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) for `GITHUB_TOKEN` so that the workflow runs successfully. 

  - [ ] `organization`: Name of the organization producing the Code Scanning integration.

**Some general notes:**

    ```
    # documentation.
    ```


dependency-graph/dependency-review.yml
```diff
@@ -0,0 +1,11 @@
...
+on: [pull_request]
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@main
```
elireisman, line 11: TODO: before GA next week, when we plan to merge this PR, it looks from the PR template docs that we'll need to replace `main` tag with a fixed commit SHA?
1

Can you confirm this is the case? If so once the `actions/dependency-review-action` repo is released we can update to a fixed SHA/tagged release as needed 👍

dependency-graph/dependency-review.yml
```diff
@@ -0,0 +1,11 @@
+name: 'Dependency Review'
+on: [pull_request]
```
elireisman, line 2: Our Action is not quite the same as those described in the PR template under `code-scanning` (the closest match IMO) our Action needs to run against PRs on push to topic branches, not `$default_branch`.

See the staff-ship repo linked in the PR description for details, and here's an example of a staff-ship installed repo where you can see the Action at work: [link](https://github.com/github/dependency-graph-api/actions/workflows/dependency-review.yml)

dependency-graph/properties/dependency-review.properties.json
```diff
@@ -0,0 +1,17 @@
...
+      "Dependency Review",
+      "Go",
+      "Java",
+      "JavaScript",
+      "TypeScript",
+      "Python",
+      "Ruby",
+      "Actions",
+      "PHP"]
+}
```
elireisman, line 17: I added tags that are not specific to language ecosystems, since our Action again is not a 1:1 match with `code-scanning` as described in PR template. Do I need to update this?

elireisman (commented):
Generally, I need to ask reviewers if our new DR Action is best registered under `code-scanning/` as described in the PR template, or can exist on it's own since the behavior (workflow triggers etc.), purpose, and team ownership is also different.

I'm happy to do whatever makes this smooth for reviewers and maintainers, and gets our new DR Action into the Security Gallery in time for GA, just not sure where this new Action fits yet 👍 guidance welcome/encouraged!

dependency-graph/properties/dependency-review.properties.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "Dependency Review",
+    "creator": "GitHub",
```
bishal-pdMSFT, line 3: No need to have creator if you want to show up as `By GitHub`.

dependency-graph/dependency-review.yml
bishal-pdMSFT: Do you want this workflow for PR against any branch? That would be lots of run without much value. That's why we usually use `$default_branch` as that's the important branch.

dependency-graph/properties/dependency-review.properties.json
```diff
@@ -0,0 +1,17 @@
...
+    "categories": [
+      "Dependency Graph",
+      "Dependency Review",
+      "Go",
+      "Java",
+      "JavaScript",
+      "TypeScript",
+      "Python",
+      "Ruby",
+      "Actions",
```
bishal-pdMSFT, line 15: Why `Actions`?

dependency-graph/dependency-review.yml
bishal-pdMSFT: Commit SHA is not needed for first party actions. You can use a version tag though instead of `main`

dependency-graph/dependency-review.yml
```diff
@@ -0,0 +1,11 @@
...
+
+jobs:
```
bishal-pdMSFT, line 4: Please add the [permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) which this job would need.

dependency-graph/properties/dependency-review.properties.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "Dependency Review",
+    "creator": "GitHub",
+    "description": "Scans Pull Requests on each push for the introduction and/or resolution of vulnerable depdendencies to the repository",
+    "iconName": "octicon mark-github",
+    "categories": [
+      "Dependency Graph",
```
bishal-pdMSFT, line 7: I am not too sure if this starter workflow would show up in `Security` section. Not sure but I think that is hard coupled today with `Code scanning` category.
@anuragc617 @Phantsure can you please confirm. If so, can we quickly fix this in the UI to allow this to show up.

dependency-graph/properties/dependency-review.properties.json
```diff
@@ -0,0 +1,17 @@
...
+      "Dependency Review",
+      "Go",
+      "Java",
+      "JavaScript",
+      "TypeScript",
+      "Python",
+      "Ruby",
+      "Actions",
+      "PHP"]
```
bishal-pdMSFT, line 16: I assume you have covered all the languages which are supported by DR action?

dependency-graph/dependency-review.yml
```diff
@@ -0,0 +1,11 @@
+name: 'Dependency Review'
```
bishal-pdMSFT, line 1: Typically we suggest few things as comments at top of the workflow file:
- What is the value provided by this workflow
- Any pre-reqs whicih must be met for the workflow to have successful runs - think some assumptions, secrets, external entities which may be needed by Action
- Any doc/support link where user can reach out to read more

bishal-pdMSFT: The intent essentially is that user feel comfortable with the workflow. And understands enough to modify it for their use case if needed.

bishal-pdMSFT, line 1: I see a new directory has been created for `dependency-graph`. While I agree that `code-scnaning` directory does not make sense but we also would not want the directories to proliferate.
@anuragc617 @Phantsure would it be okay if we rename `code-scanning` dir to `security` and keep both together.

dependency-graph/dependency-review.yml
elireisman: Yes we want this to run against any PR that's correct. The goal is for the Action to fail when a PR is introducing new known-vulnerable package versions in manifest file changes on the repo 👍

dependency-graph/dependency-review.yml
elireisman: cc @febuiles @hmaurer @lseppala as per open-source guidelines, are we obligated to semver-tag changes to the DR Action so end users can track fixed versions, or can we rely on `main` here to propagate bugfixes/DR Action changes to all installs?

🤔 The former feels like the right play, but the latter will help us avoid a long-tail support burden of folks who aren't upgrading regularly

dependency-graph/dependency-review.yml
elireisman: My understanding as per [this](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job) and discussions with the team is, our `GITHUB_TOKEN` for post-GA installs will only require `repo: read` permissions and might be fine with the `GITHUB_TOKEN` defaults?

cc @febuiles @hmaurer to confirm ?

dependency-graph/properties/dependency-review.properties.json
elireisman: We pick up Actions dependencies declared on repos that include a `.github/workflows` dir and detect vulnerable versions, same as with other language ecosystem packages

dependency-graph/properties/dependency-review.properties.json
elireisman: Excellent, can update 👍

dependency-graph/properties/dependency-review.properties.json
elireisman: That said, I'll take another pass to see if there are more appropriate tags we should include anyway thanks for the callout 🙇

dependency-graph/dependency-review.yml
bishal-pdMSFT: It would be good practice to nonetheless put "repo: read" permission as this explicitly declares that this workflow only needs such permission. Even if org/repo admin change the default GITHUB_TOKEN to be more permissive or restrictive, the workflow will have just the right level of security.

dependency-graph/dependency-review.yml
bishal-pdMSFT: @elireisman for now keep the DR starter workflow inside `code-scanning` directory. This will allow it to get shipped. And then we can follow up renaming directory to `security` and updating the UI to work with this directory.

elireisman: cc @courtneycl @saritai for docs links 🙇

dependency-graph/properties/dependency-review.properties.json
elireisman: Got a reply below, moved the files to `code-scanning/` top-level dir tree. Let me know if there's more related changes to enable this (temporary) home for security gallery entries 👍

dependency-graph/dependency-review.yml
elireisman: thanks - just moved my action spec and props files into the `code-scanning/` and `code-scanning/properties` dirs. Let me know if there's more to do there related to this move 👍

courtneycl: Doc will live here: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement

dependency-graph/dependency-review.yml
lseppala: @elireisman as discussed, since the ref used by an Action isn't required to be immutable yet (that is, you can use a branch rather than a tag), I think we should accommodate semver _and_ also have a ref that can be updated with bugfixes. To keep it simple, we can call it `v1`

@febuiles, this will require a small amount of work in the Action repo. We'll need a _branch_ named `v1` to match `main`, which is what this Action will track.

dependency-graph/dependency-review.yml
elireisman: We only want to run against `pull_request` so never default branch, only topic branch of PRs on the repo where the Action is installed 👍

dependency-graph/dependency-review.yml
elireisman: I'll update the comment now

dependency-graph/dependency-review.yml
elireisman: OK will do! thank you!

elireisman: done 👍 thank you!

elireisman:
👋 sanity check - I'm assuming we don't have to do this? `Prior to submitting a new workflow, please apply to join the GitHub Technology Partner Program` from the PR template pre-reqs?

dependency-graph/dependency-review.yml
hmaurer: @lseppala @febuiles c.f. https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management

code-scanning/properties/dependency-review.properties.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Dependency Review",
+    "description": "Scans Pull Requests on each push for the introduction and/or resolution of vulnerable depdendencies to the repository",
+    "iconName": "octicon mark-github",
+    "categories": [
+      "Code Scanning",
```
bishal-pdMSFT, line 6: You can replaced this with "Dependency Review" which is more appropriate. I have a PR which will allow that to be processed. Will most probably merge that tomorrow my day time.

bishal-pdMSFT (approved):
LGTM. Will merge once we are ready to hsip.

code-scanning/properties/dependency-review.properties.json
```diff
@@ -0,0 +1,18 @@
...
+      "Automation",
```
anuragc617, line 7: I don't think you should have Automation as categories for this templates. Automation templates are added to the automation folder. Also, move "Dependency Review" to the start of categories list.

code-scanning/properties/dependency-review.properties.json
elireisman: Sorry - did you mean the `name: "Dependency Review"` or the `categories` tag?

In the tag case, I see other two-word tags with both caps (and Dependency Review is the feature, we captialize it in other docs)

In the "name" case, the other YAML file also names this `Dependency Review` - will all those need updates as well if we change this `name`?

basically - what exactly should I update and in which file to match up with your PR change? Also - if it's easier for you to make a push to this PR directly to make the update feel free to do so, as I may not here back before tomorrow (our target ship date!) due to async/ time zones 🙇 

thanks again for the callout!

code-scanning/properties/dependency-review.properties.json
elireisman: moved "Dependency Review" to top of `categories` array, and removed "Automation" 👍 thanks!

PR #1385, opened 2022-02-03
Revert addition of Datadog Synthetics workflow

bishal-pdMSFT:
Reverts actions/starter-workflows#1342

We will bring this workflow back with a more accurate creator name.

cc @azooinmyluggage @abirismyname

PR #1503, opened 2022-03-25
Bringing back Datadog Synthetics workflow

bishal-pdMSFT:
Reverts removal of Datadog Synthetics workflow actions/starter-workflows#1385

The only change as compared to original workflow is that a `creator` fields has been added in `properties.json` which points out to `Datadog`


Phantsure:
Can we add permissions? Otherwise looks good
```
permissions:
  contents: read
```

bishal-pdMSFT:
Good point, but to be honest I am not sure what all permissions this workflow needs. @peculiarism would be better person to answer that.
Lets de-couple that change and take separately. I will create an issue for that.

Phantsure (approved):

bishal-pdMSFT:
Issue to track adding permissions: https://github.com/actions/starter-workflows/issues/1504

PR #1527, opened 2022-04-17
Shorten Rails workflow name, in line with others

indirect:
The workflows for Ruby, RubyGem, Jekyll, and similar are all just the name of the language, package, or framework. This name change brings Rails in line with the other starters.


Phantsure (approved):
This naming seems more appropriate. Thanks pointing that.

PR #1599, opened 2022-06-07
Update workflow to use the newest version of 42Crunch REST API Static…

ak1394:
Update to the existing workflow to use the newest released version of the action.


Phantsure (approved):
Seems a simple bump

PR #1619, opened 2022-06-23
Update trivy-action to fix the performance issue

knqyf263:
This version of trivy-action fixed an issue reported by GitHub. The detail is in https://github.com/aquasecurity/trivy/issues/2357.


Phantsure (approved):
Simple bump 😃

PR #1623, opened 2022-06-28
Update the sample version of the Dependency Review action.

febuiles:
This PR bumps the version of the Dependency Review action that we suggest to customers.


Phantsure (approved):
LGTM

PR #1627, opened 2022-07-05
Update dotnet.yml

andreas-nesheim:
Updated the .NET starter workflow to use .NET 6 instead of .NET 5, since .NET 5 has reached end of support.


andreas-nesheim:
Yes, I have tested and used this for many of my other workflows.

Phantsure (approved):
Seems good. Thanks @andreas-nesheim for your contributions

rexingQ (approved):
Bfg

PR #1645, opened 2022-07-26
Add starter workflows for Pages builds and deployments

TooManyBees:
## Tasks

- [x] Should specify least priviledge [permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) for `GITHUB_TOKEN` so that the workflow runs successfully.

**Some general notes:**

    ```
    # documentation.
    ```


bishal-pdMSFT:
@anuragc617 these workflows in `Pages` folder will not show up automatically in Actions new page as we hard-code the allowed categories in UI. Right?

pages/gatsby.yml
```diff
@@ -0,0 +1,96 @@
...
+        id: detect-package-manager
+        run: |
+          if [ -f "${{ github.workspace }}/yarn.lock" ]; then
+            echo "::set-output name=manager::yarn"
+            echo "::set-output name=command::install"
+            exit 0
+          elif [ -f "${{ github.workspace }}/package.json" ]; then
+            echo "::set-output name=manager::npm"
+            echo "::set-output name=command::ci"
+            exit 0
```
bishal-pdMSFT, line 48: Not for this PR but there exists another way to detect package managers which can be used by UI to offer simpler workflows. We can sync up separately in this.

pages/jekyll-gh-pages.yml
```diff
@@ -0,0 +1,50 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment
```
bishal-pdMSFT, line 18: Neat. @anuragc617 @Phantsure Probably something we should consider doing for other deployment workflows as well.

pages/properties/static.properties.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Static HTML",
+    "creator": "GitHub Actions",
+    "description": "Deploy static files in a repository without a build.",
+    "iconName": "html",
+    "categories": ["Pages"]
```
anuragc617, line 6: We should also include language/frameworks targeted by this template in categories. It will be helpful if we use our suggestion engine for Pages in future.

pages/properties/nuxtjs.properties.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "NuxtJS",
+    "creator": "GitHub Actions",
+    "description": "Package a NuxtJS site.",
+    "iconName": "nuxtjs",
+    "categories": ["Pages"]
```
anuragc617, line 6: We should also include language/frameworks targeted by this template in categories. It will be helpful if we use our suggestion engine for Pages in future. Add `nuxtjs` to categories.

pages/properties/nextjs.properties.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Next.js",
+    "creator": "GitHub Actions",
+    "description": "Package a Next.js site.",
+    "iconName": "nextjs",
+    "categories": ["Pages"]
```
anuragc617, line 6: Add nextjs to the categories.

pages/properties/static.properties.json
bishal-pdMSFT: @anuragc617 I don't Pages UI is taking advantage of this today. Hence we can proceed with PR and take these separately.

pages/properties/gatsby.properties.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Gatsby",
+    "creator": "GitHub Actions",
```
Phantsure, line 3: We don't really need the `creator` if GitHub is the creator. That is set by default.

anuragc617 (approved):

PR #1723, opened 2022-08-30
use latest version of Hugo

coliff:
REF: https://github.com/gohugoio/hugo/releases

### DIFF
--- a/pages/hugo.yml
+++ b/pages/hugo.yml
@@ -30,7 +30,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      HUGO_VERSION: 0.99.0
+      HUGO_VERSION: 0.102.1
     steps:
       - name: Install Hugo CLI
         run: |
--- a/pages/hugo.yml
+++ b/pages/hugo.yml
@@ -30,7 +30,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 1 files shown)


Phantsure:
@yoannchaudet Does it make sense to keep `Hugo` up to date always?

coliff:
> Does it make sense to keep `Hugo` up to date always?

Occasionally there are breaking changes from release to release in Hugo so I don't think using 'latest' by default would be a good idea... but I would like there to be an option to use 'latest'.

pages/hugo.yml
```diff
@@ -30,7 +30,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      HUGO_VERSION: 0.99.0
+      HUGO_VERSION: 0.102.1
```
JamesMGreene, line 33: There have already been 2 additional patch releases since this PR was created. 😅 

```suggestion
      HUGO_VERSION: 0.102.3
```

JamesMGreene (approved):
Verified with our internal Pages regression testing repo for Hugo. 👍🏻

JamesMGreene:
> but I would like there to be an option to use 'latest'.

@coliff Just remember that these are also just starter workflows. Once they've landed in your repository, you can always update that instance of them yourself as well. 😊

PR #1724, opened 2022-08-30
branches: [ `$default-branch` ]

austenstone:
This regex `^(?!.*(default-branch)).*branches:.*` should help find anything not using [$default-branch](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/).

### DIFF
--- a/.github/workflows/sync-ghes.yaml
+++ b/.github/workflows/sync-ghes.yaml
@@ -2,8 +2,7 @@ name: Sync workflows for GHES
 
 on:
   push:
-    branches:
-    - main
+    branches: [ $default-branch ]
 
 jobs:
   sync:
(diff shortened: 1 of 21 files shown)


Phantsure:
@austenstone What is the reasoning behind this change?

austenstone:
> @austenstone What is the reasoning behind this change?

I was researching the $default-branch macro and figured I'd update the starter workflows based on the blog post

> We have updated all of our [starter workflows](https://github.com/actions/starter-workflows) to use a new $default-branch macro rather than the previously hardcoded master.

lukins (commented):
Skvělé

Phantsure (approved):
Change looks aligned with the blog referred. We should have this in guidelines as well for future actions.

PR #1757, opened 2022-09-21
Hide ESLint on GHES

simon-engledew:
👋 Hi! I'm from the GitHub Code Scanning team.

We've noticed that ESLint is being incorrectly displayed on GHES.

This change should hide the workflow. 🙇


nickfyson (approved):
LGTM

PR #1759, opened 2022-09-21
Make enterprise opt-in instead of opt-out

simon-engledew:
This should stop workflows appearing on GHES when they have not explicitly been vetted as suitable.


vilacides (approved):

nickfyson (approved):
LGTM

Phantsure:
@simon-engledew Do you have a process for vetting code-scanning templates?

PR #1760, opened 2022-09-21
fix trigger for ghes sync

nickfyson:
I think this bug snuck in when [updating all the template workflows to use `$default_branch`](https://github.com/actions/starter-workflows/pull/1724).

### DIFF
--- a/.github/workflows/sync-ghes.yaml
+++ b/.github/workflows/sync-ghes.yaml
@@ -2,7 +2,7 @@ name: Sync workflows for GHES
 
 on:
   push:
-    branches: [ $default-branch ]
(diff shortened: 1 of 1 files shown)


Phantsure:
Is the default branch not `main`?

Phantsure (approved):

Phantsure:
Interesting. It seems to fix that

nickfyson:
The mistake was in accidentally applying the use of a template variable inside an actual workflow. Rendering of the templates happens in the setup experience, the string `$default_branch` doesn't mean anything to Actions itself...  🙂

Phantsure:
Thanks for that

PR #1762, opened 2022-09-22
Add pages directory to readme

nickmccurdy:

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,10 +12,11 @@ These are the workflow files for helping people get started with GitHub Actions.
 
 ### Directory structure
 
-* [ci](ci): solutions for Continuous Integration workflows.
-* [deployments](deployments): solutions for Deployment workflows.
-* [automation](automation): solutions for automating workflows.
-* [code-scanning](code-scanning): starter workflows for [Code Scanning](https://github.com/features/security)
+* [ci](ci): solutions for Continuous Integration workflows
+* [deployments](deployments): solutions for Deployment workflows
+* [automation](automation): solutions for automating workflows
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +41,14 @@ For example: `ci/django.yml` and `ci/properties/django.properties.json`.
...
 * Automation
 * utilities
+* Pages
+* Hugo
+* Gatsby
+* Next
+* Nuxt
+* HTML
+* JavaScript
+* CSS
```
Phantsure, line 51: I think we don't need these languages/frameworks in categories. `Pages` seems enough

Phantsure: These languages/frameworks are covered in `Valid properties` section under categories

nickmccurdy: Now only the frameworks are there, since they're still in the properties files.

Phantsure: I think frameworks are also not required. Pages should be enough

nickmccurdy: Wouldn't it be better to have this consistent with the properties files? Since languages aren't mentioned there IIRC.

Phantsure: Kindly only keep `Hugo` as that is the only one not mentioned here: https://github.com/github-starter-workflows/repo-analysis-partner/blob/main/tech_stacks.yml

Phantsure (approved):

stevenjoezhang:
@nickmccurdy @Phantsure There seems to be an extra `* Hugo` under `* Pages` in README.md

nickmccurdy:
We intentionally include Hugo as an extra category as it's not mentioned anywhere else.

Source: https://github.com/actions/starter-workflows/pull/1762#discussion_r996963757

PR #1775, opened 2022-10-07
code-scanning/scorecards.yml: bump action versions/hashes, clarify comments, comment out private-repo-only permissions

pnacht:
Closes ossf/scorecard-action#945 and ossf/scorecard-action#946.

- Bumps all actions, including scorecards-action from 2.0.3 to 2.0.4
- Expands on comments regarding purpose of different triggers and permissions, clarifies comments regarding `publish_results`.
- Comments out the permissions which are only necessary in private repos (`contents` and `actions`). This is consistent with the `repo_token` argument, which is only already commented out since it's mostly only used for private repositories.
- Adds a disclaimer to the top of the workflow as requested by the PR template.

cc: @azeemshaikh38


NostalgicGareth (approved):
- [ ]

code-scanning/scorecards.yml
```diff
@@ -17,20 +24,20 @@ jobs:
...
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # tag=v3.0.0
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
         with:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@865b4092859256271290c77adbd10a43f4779972 # tag=v2.0.3
+        uses: ossf/scorecard-action@e363bfca00e752f91de7b7d2a77340e2e523cb18 # v2.0.4
```
Phantsure, line 40: All looks good. We can merge this but would you like to refer to more latest release?

Phantsure (approved):

code-scanning/scorecards.yml
pnacht: Good catch! Hadn't noticed the new releases. Updated.

Issue #1754, opened 2022-09-20
Missing Link from python-package.yml


John-Poulos:
Line #2 in python-package.yml at https://github.com/actions/starter-workflows/blob/main/ci/python-package.yml
appears to be a dead link. It will bring up the generic support page (https://support.github.com/) instead.

From yaml: 
# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions

chabala:
Indeed, all the deeplinks in comments seem to be broken, e.g. in `maven.yml` https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven also redirects to the generic support page.

PR #1777, opened 2022-10-10
Update old documentation links

Phantsure:
Addresses issue https://github.com/actions/starter-workflows/issues/1754


ci/swift.yml
```diff
@@ -1,3 +1,6 @@
+# This workflow will build a golang project
```
ScottBrenner, line 1:
```suggestion
# This workflow will build a Swift project
```

ScottBrenner (changes requested):

ci/swift.yml
Phantsure: Thanks. Good catch

anuragc617 (approved):

PR #1806, opened 2022-11-04
Remove extra whitespaces

jorgectf:
This PR fixes extra whitespaces for `yaml` files. This is just a temporary fix tho, the best approach would be to set up a [pre-commit hook](https://github.com/pre-commit/pre-commit-hooks#pre-commit-hooks) or PR checks to improve the workflow set up experience for users.

PR #1823, opened 2022-11-16
Add Pull Request check

jorgectf:
This PR adds [a workflow to check](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#checks) that Pull Requests are correctly formatted, currently using **only** the `trailing-whitespace` script. For more information, check the [available pre-commit hooks](https://pre-commit.com/hooks.html).

### Context

[This PR](https://github.com/actions/starter-workflows/pull/1806) fixed most of the files using this script. During an internal activity, we found that, if a user sets up a workflow via GitHub UI and the repository has these kind of checks enabled, the user would need to fix them. This could mislead the user or discourage them to set up the workflow.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,32 @@
+name: Lint
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+
+  pre-commit:
+    name: pre-commit
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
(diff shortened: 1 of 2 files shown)


jorgectf:
cc @Phantsure

Phantsure:
Thanks a lot @jorgectf. Would need to fix those changes suggested by linter first. And can we keep it to only template folders and icons?

Phantsure (approved):

jorgectf:
Oh, this slipped through the cracks, sorry @Phantsure! Thanks for the changes 👍

PR #1844, opened 2022-11-30
Fix whitespace lint

Phantsure:
Update lint workflow. This workflow would report extra white spaces in different files


.github/workflows/lint.yaml
```diff
@@ -1,7 +1,7 @@
 name: Lint
 
 on:
-  pull_request:
+  pull_request_target:
```
anuragc617, line 4: why do we need this to be pull_request_target?

Phantsure: So that they don't require approval for seeing linting  issue

anuragc617: For forked repo PRs, workflows with `pull_request` get repo `read` permission while workflows triggered with `pull_request_target` get repo `read/write`  permission. The `pull_request_target` should not be used  for workflows that can commit the code repo.  Is this workflow only points out linting issues or it fixes them as well via code commit? If it is pushing code then we should not use `pull_request_target`.

Phantsure: It fixes them as well but doesn't commit or push the changes. Changes are made to local runner git repo

Phantsure: Have made it back to `pull_request`

Phantsure: back to `pull_request_target`

PR #1850, opened 2022-12-08
Fixed misspelling of "privileged"

norwd:
I noticed this while filling out the PR description for #1849.


Phantsure (approved):
Thank you for that

PR #1855, opened 2022-12-12
Fix and improve Hugo starter action

### LINKED ISSUE #1854: Fix and improve Hugo starter action
Fix:

- The Hugo download URL is no longer valid

Improve:

- Update Hugo to version to 0.108.0
- Install Dart Sass transpiler (site authors are transitioning from LibSass)
- Install Node.js dependencies (typically postcss, autoprefixer, and purgeccs)

I will submit a PR.

jmooring:
Closes #1854

### DIFF
--- a/pages/hugo.yml
+++ b/pages/hugo.yml
@@ -30,12 +30,14 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      HUGO_VERSION: 0.102.3
+      HUGO_VERSION: 0.108.0
     steps:
       - name: Install Hugo CLI
         run: |
-          wget -O ${{ runner.temp }}/hugo.deb https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_extended_${HUGO_VERSION}_Linux-64bit.deb \
+          wget -O ${{ runner.temp }}/hugo.deb https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_extended_${HUGO_VERSION}_linux-amd64.deb \
           && sudo dpkg -i ${{ runner.temp }}/hugo.deb
+      - name: Install Dart Sass Embedded
+        run: sudo snap install dart-sass-embedded
       - name: Checkout
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


DiqahLer (approved):

pages/hugo.yml
```diff
@@ -30,19 +30,28 @@ jobs:
...
-      HUGO_VERSION: 0.102.3
+      HUGO_VERSION: 0.108.0
+      EMBEDDED_DART_SASS_VERSION: 1.56.2
     steps:
       - name: Install Hugo CLI
         run: |
-          wget -O ${{ runner.temp }}/hugo.deb https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_extended_${HUGO_VERSION}_Linux-64bit.deb \
+          wget -O ${{ runner.temp }}/hugo.deb https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_extended_${HUGO_VERSION}_linux-amd64.deb \
           && sudo dpkg -i ${{ runner.temp }}/hugo.deb
+      - name: Install Dart Sass Embedded
```
yoannchaudet, line 40: 👋 Hello and thank you for your contribution.

Could we make this step optional if `EMBEDDED_DART_SASS_VERSION` is empty (and add a comment about that line 34)?

I am not sure how Dart Sass Embedded is popular nowadays with Hugo but making the starter workflow as simple to customize as possible for people that may not be too familiar with Actions is something important.

jmooring: > I am not sure how Dart Sass Embedded is popular nowadays with Hugo

With LibSass now deprecated, usage will continue to increase. Over time, independent theme creators will update their themes to use the Dart Sass transpiler. So a new user, making use of a prebuilt theme, will not have a good experience.

I will make the step optional if EMBEDDED_DART_SASS_VERSION is empty, but it should not empty by default. This should just... work.

jmooring:
Also, if you would prefer, we could install dart-sass-embedded via `sudo snap install dart-sass-embedded`. This snap is automatically updated, so no need to specify version.

yoannchaudet (approved):

PR #1944, opened 2023-02-16
Update scorecard.yml with latest releases

ChrisCarini:
Update scorecard.yml with latest releases for ossf/scorecard-action & github/codeql-action/upload-sarif

Noticed these were out-of-date when I saw the below warning on a recent run ([ref](https://github.com/ChrisCarini/automatic-github-issue-navigation-configuration-jetbrains-plugin/actions/runs/4190143423/jobs/7263241173)):
Screenshot 2023-02-15 at 19 09 45

PR #1951, opened 2023-02-19
 "Update scorecard.yml with latest releases"

emmyforeignet:
actions/starter-workflows#1944


emmyforeignet:
@emmyforeignet:revert-1944-patch-1

Phantsure:
@emmyforeignet What is the reason for this?

PR #2033, opened 2023-04-26
Update elixir.yml - fix default build and bump version

stepchud:
Default builds on skeleton Elixir projects are failing with this error:
```
[Build and test]
Tried to map a target OS from env. variable 'ImageOS' (got ubuntu22), but failed. If you're using a self-hosted runner, you should set 'env': 'ImageOS': ... to one of the following: ['ubuntu16', 'ubuntu18', 'ubuntu20', 'win16', 'win19']
```
The error message is misleading because this happened while using the default workflow elixir.yml setup recommended by GitHub for Elixir. This PR fixes the failing build and also bumps the Erlang/Elixir versions to the most recent releases.

The issue was originally reported here: https://github.com/erlef/setup-beam/issues/188#issue-1684162372, let me know if I should make an Issue on this repo.

## Pre-requisites

- [x] existing starter workflow

---

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```


ci/elixir.yml
```diff
@@ -23,10 +23,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Set up Elixir
-      uses: erlef/setup-beam@988e02bfe678367a02564f65ca2e37726dc0268f
+      uses: erlef/setup-beam@v1
```
JamesMGreene, line 26: Although not the case with _every_ action, we prefer to use pinned commit SHA for improved security:

```suggestion
      uses: erlef/setup-beam@61e01a43a562a89bfc54c7f9a378ff67b03e4a21 # v1.16.0
```

ci/elixir.yml
```diff
@@ -23,10 +23,10 @@ jobs:
...
       with:
-        elixir-version: '1.12.3' # Define the elixir version [required]
-        otp-version: '24.1' # Define the OTP version [required]
+        elixir-version: '1.14.4' # Define the elixir version [required]
+        otp-version: '25.3' # Define the OTP version [required]
```
JamesMGreene, line 29: Updated to latest:

```suggestion
        elixir-version: '1.15.2' # [Required] Define the Elixir version
        otp-version: '26.0'      # [Required] Define the Erlang/OTP version
```

JamesMGreene (approved):
Thanks for the bump. I've bumped them further to the latest versions given this PR has been sitting for a while. 💝

PR #2091, opened 2023-06-12
CodeQL: Update autobuild comment

henrymercer:
Previously Swift was not included in the list of languages that autobuild will try to build.  This could lead to customers inadvertently removing the autobuild step if they're running Swift because the comment suggests it will have no effect.

This PR adds Swift to this list, to avoid this potential confusion.  Since it only updates a comment, this is a very safe change.


965022-oss:
Merged #2628 
Branch: main
965022-oss-patch3
SHA: f4dbb8424c

PR #2133, opened 2023-08-01
Split cmake starter workflow into single/multi platform support

jgoshi:
Rename the existing starter workflow to indicate that it supports a single platform. Add a new starter workflow that has cross-platform support.

## Pre-requisites

---

---

## Tasks

- [NA] Packaging workflows should run on `release` with `types: [ created ]`.
- [NA] Publishing workflows should have a filename that is the name of the language or platform, in lower case, followed by "-publish" (for example, [`docker-publish.yml`](https://github.com/actions/starter-workflows/blob/main/ci/docker-publish.yml)).

- [NA] Should be preserved under [the `code-scanning` directory](https://github.com/actions/starter-workflows/tree/main/code-scanning).
- [NA] Should include a matching `code-scanning/properties/*.properties.json` file (for example, [`code-scanning/properties/codeql.properties.json`](https://github.com/actions/starter-workflows/blob/main/code-scanning/properties/codeql.properties.json)), with properties set as follows:
  - [NA] `name`: Name of the Code Scanning integration.
  - [NA] `creator`: Name of the organization/user producing the Code Scanning integration.
  - [NA] `description`: Short description of the Code Scanning integration.
  - [NA] `categories`: Array of languages supported by the Code Scanning integration.
  - [NA] `iconName`: Name of the SVG logo representing the Code Scanning integration. This SVG logo must be present in [the `icons` directory](https://github.com/actions/starter-workflows/tree/main/icons).
- [NA] Should run on `push` to `branches: [ $default-branch, $protected-branches ]` and `pull_request` to `branches: [ $default-branch ]`. We also recommend a `schedule` trigger of `cron: $cron-weekly` (for example, [`codeql.yml`](https://github.com/actions/starter-workflows/blob/c59b62dee0eae1f9f368b7011cf05c2fc42cf084/code-scanning/codeql.yml#L14-L21)).

**Some general notes:**

    ```
    # documentation.
    ```


ci/cmake-single-platform.yml
michaelbprice, line 1: I'm wondering if we shouldn't be more explicit and name this `cmake-ubuntu.yml`. I guess a possible determining factor would be the answer to:

How easily could take this single platform script and convert it to run on a different platform?

ci/cmake-single-platform.yml
```diff
@@ -1,4 +1,6 @@
-name: CMake
+# This starter workflow is for a CMake project running on a single platform. There is a different starter workflow if you need cross-platform coverage.
+# See: https://github.com/actions/starter-workflows/blob/main/ci/cmake-multi-platform.yml
+name: CMake on a single platform
```
michaelbprice, line 3: Along the same lines as the naming issue, maybe just `CMake on Ubuntu`?  There's a tradeoff:

* `CMake on Ubuntu` - The user gets a clear picture from the GitHub.com UI that the workflow is for Ubuntu
VS
* `CMake on a single platform` - The user won't need to make changes to the name if they change the nature of the workflow (or copy it to make a separate Windows or Mac workflow file). This also avoids the risk of having an "incorrect" name.

ci/properties/cmake-multi-platform.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "CMake based multi-platform projects",
```
michaelbprice, line 2:
```suggestion
    "name": "CMake based, multi-platform projects",
```

ci/properties/cmake-single-platform.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "CMake based single-platform projects",
```
michaelbprice, line 2:
```suggestion
    "name": "CMake based, single-platform projects",
```

ci/properties/cmake-single-platform.json
michaelbprice, line 1: This file and its contents are also subject to the naming question.

ci/properties/cmake-multi-platform.json
```diff
@@ -0,0 +1,6 @@
...
+    "description": "Build and test a CMake based project on multiple platforms.",
+    "iconName": "cmake",
```
michaelbprice, line 4: @andrewakim Do we know what this `iconName` field controls? If it's used to select an icon, should we consider having some sort of adorned icon to distinguish them?

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,60 @@
...
+          - os: windows-latest
+            compiler: clang
+          - os: ubuntu-latest
+            compiler: cl
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set reusable strings
+      # Turn repeated input strings (such as the build output directory) into step outputs. These step outputs can be used throughout the workflow file.
```
michaelbprice, line 41: Is this style of putting the comment after the first field of the step consistent with other workflow files? It feels strange to me personally.

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,60 @@
...
+    branches: [ $default-branch ]
+  pull_request:
+    branches: [ $default-branch ]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      # Set fail-fast to false to ensure that feedback is delivered for all matrix combinations. Can consider changing this to true when the workflow is stable.
```
michaelbprice, line 16: Should we use column-limit for the contents of the file, particularly for comments and for longer commands?

@andrewakim - Do you know if there's a GitHub Actions YAML formatting action that could just apply some standards across starter workflow file contents (and other workflows if others wanted to use it).

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set reusable strings
+      # Turn repeated input strings (such as the build output directory) into step outputs. These step outputs can be used throughout the workflow file.
+      id: strings
+      shell: bash
+      run: |
+        echo "build-output-dir=${{github.workspace}}/build" >> "$GITHUB_OUTPUT"
```
michaelbprice, line 45: One of the things that I've noticed within GitHub Actions documentation is that workflow expressions are usually formatted a particular way, namely instead of
```
${{github.workspace}}
```
You see (with leading and trailing whitespace)
```
${{ github.workspace }}
```

@andrewakim - Is this a style thing that we should reproduce here as well?

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    - name: Configure CMake
+      # Configure CMake in a 'build' subdirectory. `CMAKE_BUILD_TYPE` is only required if you are using a single-configuration generator such as make.
+      # See https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html?highlight=cmake_build_type
+      run: cmake -B ${{ steps.strings.outputs.build-output-dir }} -DCMAKE_CXX_COMPILER=${{matrix.compiler}} -DCMAKE_C_COMPILER=${{matrix.compiler}} -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -S ${{github.workspace}}
```
michaelbprice, line 50: Even if we don't enforce a column width limit, I think we should make this a multi-line run step. It's the one that users are most likely to modify/extend (an assumption I'm making), so it's the one they'll most likely be tempted to make multi-line, which if you don't already know how, can be frustrating.

```suggestion
      run: |
        cmake -B ${{ steps.strings.outputs.build-output-dir }} \
        -DCMAKE_CXX_COMPILER=${{matrix.compiler}} \
        -DCMAKE_C_COMPILER=${{matrix.compiler}} \
        -DCMAKE_BUILD_TYPE=${{matrix.build_type}} \
        -S ${{github.workspace}}
```

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    - name: Build
+      # Build your program with the given configuration. Note that --config is needed because the default Windows generator is a multi-config generator (Visual Studio generator).
+      run: cmake --build ${{ steps.strings.outputs.build-output-dir }} --config ${{matrix.build_type}}
+
+    - name: Test
+      working-directory: ${{ steps.strings.outputs.build-output-dir }}
+      # Execute tests defined by the CMake configuration. Note that -C is needed because the default Windows generator is a multi-config generator (Visual Studio generator).
+      # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
+      run: ctest -C ${{matrix.build_type}}
```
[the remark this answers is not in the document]
michaelbprice: We could probably do similar things to the contents of the single platform workflow, but let's avoid that for now so that we can avoid having multiple variables while we try to understand if users are finding the multi-platform workflow a better experience.

ci/cmake-multi-platform.yml
michaelbprice, line 50: `-DCMAKE_CXX_COMPILER=${{matrix.compiler}} -DCMAKE_C_COMPILER=${{matrix.compiler}}` is not quite correct. For `cl` it is, but for GCC and Clang, the correct value to use for `CMAKE_CXX_COMPILER` are the `g++` and `clang++` frontend driver programs, repsectively.

michaelbprice (changes requested):
I would like to see at least 1, perhaps 2, examples of this workflow applied to and working well for some example C++ code projects. If you need suggestions on ones to try, I can hunt some down. That would also let us fork and experiment with what the experience is to take these workflows and extend them for additional cases (like adding Mac support, adding in a dependency management step, or moving to Ninja as a generator).

ci/cmake-single-platform.yml
jgoshi: I think it's straightforward to just change the os. I think maybe we can leave the name for now?

ci/cmake-multi-platform.yml
jgoshi: You mean in the matrix I should use g++ and clang++ instead of gcc and clang?

michaelbprice: More complex than that. An example invocation of cmake should look like this:
```
cmake -B ./build -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc -DCMAKE_BUILD_TYPE=Release -S .
```

michaelbprice: The problem is that the `g++` driver and the `gcc` driver might have different built-in include paths to search, which could cause #include directive failures if `gcc` is used when `g++` should be. I think there are also some linker implications as well (i.e., how the driver invokes `ld`).

ci/cmake-single-platform.yml
jgoshi: Checked with Michael offline. We'll keep the naming convention for now. He said the good thing is that with different names, we should be able to see what mix of platforms people are targeted based on the file name, so in theory we can make a data argument to support any further change (or not).

ci/cmake-single-platform.yml
jgoshi: Keeping the naming convention as is for now (see the other comment for the reasoning).

ci/properties/cmake-multi-platform.json
jgoshi: Will make this change and push the updates.

ci/properties/cmake-single-platform.json
jgoshi: Same as above.

ci/properties/cmake-single-platform.json
jgoshi: Keeping as is for now.

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,75 @@
+# This starter workflow is for a CMake project running on multiple platforms. There is a different starter workflow if you just want a single platform.
+# See: https://github.com/actions/starter-workflows/blob/main/ci/cmake-single-platform.yml
+name: CMake on a multiple platforms
```
michaelbprice, line 3:
```suggestion
name: CMake on multiple platforms
```

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,75 @@
...
+    branches: [ $default-branch ]
+  pull_request:
+    branches: [ $default-branch ]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      # Set fail-fast to false to ensure that feedback is delivered for all matrix combinations. Can consider changing this to true when the workflow is stable.
```
michaelbprice, line 16:
```suggestion
      # Set fail-fast to false to ensure that feedback is delivered for all matrix combinations. Consider changing this to true when your workflow is stable. 
```

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,75 @@
...
+          - os: ubuntu-latest
+            c_compiler: clang
+            cpp_compiler: clang++
+        exclude:
+          - os: windows-latest
+            c_compiler: gcc
+          - os: windows-latest
+            c_compiler: clang
+          - os: ubuntu-latest
+            c_compiler: cl
```
michaelbprice, line 45: Since you build all of the relevant matrix entries in the `include`, I think you can make this simplification as well. Obviously, you should test this suggestion.

```suggestion
      matrix:
        os: [ubuntu-latest, windows-latest]
        build_type: [Release]
        include:
          - os: windows-latest
            c_compiler: cl
            cpp_compiler: cl
          - os: ubuntu-latest
            c_compiler: gcc
            cpp_compiler: g++
          - os: ubuntu-latest
            c_compiler: clang
            cpp_compiler: clang++
```

jgoshi: Unfortunately, this doesn't work because include will only add a new entry if it doesn't overwrite something already added by another include. So the 2 ubuntu ones only one will apply (because only "ubuntu-latest" comes from the matrix outside the include statement)

michaelbprice: Yeah, you're right. I think there could be a way to make it work, but it would be really weird and probably not worth the trouble. In any manner, having an `include` and `exclude` block isn't the end of the world as it serves as a good example of how to construct that sparse matrix.

jtamsut (commented):
@jgoshi We have a [PR template](https://github.com/actions/starter-workflows/blob/main/.github/pull_request_template.md) with a checklist for all new contributions to this repo. Can you go through and make sure all (relevant) items of the checklist are completed? Can you also use the PR template in your PR description? Thanks

ci/properties/cmake-multi-platform.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "CMake based, multi-platform projects",
+    "description": "Build and test a CMake based project on multiple platforms.",
+    "iconName": "cmake",
+    "categories": ["Continuous integration", "C", "C++"]
+}
```
jtamsut, line 6: nit: Add newline to end of file

ci/properties/cmake-single-platform.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "CMake based, single-platform projects",
+    "description": "Build and test a CMake based project on a single-platform.",
+    "iconName": "cmake",
+    "categories": ["Continuous integration", "C", "C++"]
+}
```
jtamsut, line 6: same nit as above: add a newline

ci/properties/cmake-multi-platform.json
jtamsut: From the [GitHub docs](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization#creating-a-starter-workflow):

> iconName - Optional. Specifies an icon for the workflow that is displayed in the list of workflows. iconName can one of the following types:
>
>    - An SVG file that is stored in the workflow-templates directory. To reference a file, the value must be the file name without the file extension. For example, an SVG file named example-icon.svg is referenced as example-icon.
> 
>    - An icon from GitHub's set of [Octicons](https://primer.style/octicons/). To reference an octicon, the value must be octicon <icon name>. For example, octicon smiley.

ci/cmake-multi-platform.yml
jtamsut: Looking through the other workflows the convention that seems to be applied here is to add spaces before the opening and closing curly braces:

```
${{ stuff }}
```

ci/properties/cmake.properties.json
jtamsut, line 1: Why did you delete this file? According to the [docs](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization#creating-a-starter-workflow) you need this. Am I missing something?

ci/cmake-multi-platform.yml
jgoshi: Made this change to add spaces.

ci/properties/cmake-multi-platform.json
jgoshi: Fixed it

ci/properties/cmake-single-platform.json
jgoshi: Fixed it.

jgoshi:
@jtamsut I went through the PR template checklist and used it in the description. Is this what you meant? Also regarding joining the GitHub Technology Partner Program I didn't do that part but work for Microsoft who I assume is a partner. Or do I need to join individually?

ci/properties/cmake-multi-platform.json
michaelbprice: Thanks @jtamsut. I went looking to see if there were any public domain icons that could help distinguish these workflows, but I couldn't really find any. I did find that the current CMake icon we are using has been updated, but [the version available on Wikimedia Commons](https://commons.wikimedia.org/wiki/File:Cmake.svg) is only 200 x 200. I think we are fine skipping this for now.

michaelbprice (approved):
I'm approving. Pending a validation that it works on a fairly simple, but larger project, I don't think I have any further input.

I'll be away next week, so as soon as we get the required approvals from GitHub folks, I have no objections to a merge.

ci/cmake-multi-platform.yml
```diff
@@ -0,0 +1,75 @@
...
+        -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+        -S ${{ github.workspace }}
+
+    - name: Build
+      # Build your program with the given configuration. Note that --config is needed because the default Windows generator is a multi-config generator (Visual Studio generator).
+      run: cmake --build ${{ steps.strings.outputs.build-output-dir }} --config ${{ matrix.build_type }}
+
+    - name: Test
+      working-directory: ${{ steps.strings.outputs.build-output-dir }}
+      # Execute tests defined by the CMake configuration. Note that -C is needed because the default Windows generator is a multi-config generator (Visual Studio generator).
```
LukaszMendakiewicz, line 73: Should `-C` be `--build-config` in this comment?

LukaszMendakiewicz: And in single-platform, should the `-C` parameter be changed as well for consistency?

LukaszMendakiewicz (approved):

ci/cmake-multi-platform.yml
jgoshi: Thanks. I'll fix the comment. Also for the single-platform one, leaving it for now. This was what Michael had said (in another comment on this PR):
"We could probably do similar things to the contents of the single platform workflow, but let's avoid that for now so that we can avoid having multiple variables while we try to understand if users are finding the multi-platform workflow a better experience."

jgoshi:
For validation I run both the single-platform and multi-platform workflows on a simple CMake project with a single test. I also ran the multi-platform workflow on a fork of the https://github.com/llvm/llvm-project repo (to build Clang). The Windows one passed, but both Ubuntu ones failed exceeding the max timeout of 360 minutes (both were still building). But things looked like they were working.

Per Michael's earlier comment "I'll be away next week, so as soon as we get the required approvals from GitHub folks, I have no objections to a merge". What are the next steps here in order to merge this? Thank you.

ci/cmake-multi-platform.yml
andrewakim: Hmmm. I don't think so and I'm not sure what we enforced in the past.

andrewakim (approved):

jgoshi:
@andrewakim Do you know what the next steps are? It says "At least 1 approving review is required by reviewers with write access".

jtamsut:
> @jtamsut I went through the PR template checklist and used it in the description. Is this what you meant? Also regarding joining the GitHub Technology Partner Program I didn't do that part but work for Microsoft who I assume is a partner. Or do I need to join individually?

Hey @jgoshi. Yeah since you are from Microsoft you don't need to become a Technology Partner. You can skip that prereq.

jtamsut:
@jgoshi I see you listed this step as NA:

[NA] This workflow must only use actions that are produced by the language or ecosystem that the workflow supports. These actions must be [published to the GitHub Marketplace](https://github.com/marketplace?type=actions). We require that these actions be referenced using the full 40 character hash of the action's commit instead of a tag. Additionally, workflows must include the following comment at the top of the workflow file:

```
# documentation.
```

Why? Can you add the blurb above to the top of the workflow file?

jgoshi:
Sorry I put N/A by mistake. It only uses the checkout action so I'll change the checklist.

jtamsut (approved):

965022-oss:
I was newly starter workflows for Code Scanning follow up existing update.

ci/cmake-multi-platform.yml
swapnilrajegithubcom: You mean the should new project creation

965022-oss: It's a new action/workflow cmake-multi-platform.yml, created build type configuration using CMAKE_BUILD_TYPE now I just have to configured variables able cmake configuration connected on data application.

ci/cmake-multi-platform.yml
965022-oss: Workflow run with updated 
strings

PR #2189, opened 2023-10-13
Update jekyll-gh-pages.yml

fulldecent:
Bump dep


JamesMGreene:
Updated separately in:
- https://github.com/actions/starter-workflows/pull/2244

Thank you, though! ❤️

PR #2226, opened 2023-11-19
chore: update node versions based on releases

iennae:
Current Active/Maintenance are Node.js 16, 18, 20. (21 is an odd release so I don't think should be included.) 

Also updating the release link for documentation as the prior one is invalid.

This is just an update to an existing action and doesn't modify the workflow.


iennae:
at this point #2269 replaces this one based on the current status of supported node versions.

PR #2256, opened 2023-12-12
Pages: remove next export step in nextjs.yml

xvvhang:
The "next export" command has been removed in favor of "output: export" in next.config.js. So there's no need to run next export here any more.


JamesMGreene:
@xvvhang Thanks for the PR!

This should be resolved now if you create a new Next.js starter workflow using our updated template. 🚀 

Upgrading to use the latest `actions/configure-pages@v5` also offers better static site generator configuration support for Next.js >= 13.3.0.

Issue #2159, opened 2023-09-17
ci/rubyonrails.yml bug


jamiemccarthy:
## Issue

I believe there's a bug in `ci/rubyonrails.yml`.

In [lines 48-54](https://github.com/indirect/starter-workflows/blob/1dc689b281a29f0f4baaea4faf2783412735b44e/ci/rubyonrails.yml#L48-L54), the lint job sets up ruby with `ruby/setup-ruby` and `bundler-cache: true`. The next step attempts to run `bin/bundler-audit`.

In a stock configuration the working directory is `/home/runner/work/REPO_NAME/REPO_NAME` and there's no `bin` subdirectory. That always fails. (For the audit and for the other two CI steps as well.)

The `bundler-audit` binary is installed in `/home/runner/work/REPO_NAME/REPO_NAME/vendor/bundle/ruby/RUBY_VERSION/bin/`, and unless commands are run with `bundle exec`, that directory isn't in the `$PATH`.

The fix I would suggest is

1. to run the three linting commands with `bundle exec` as a prefix (and not specifying the `bin/` directory, that's inappropriate).

Other changes I would suggest:

2. spell it `bundle-audit`, not `bundler-audit`, as [suggested in its readme](https://github.com/rubysec/bundler-audit#readme)
3. run `ruby/setup-ruby` at version `v1`, not pinned to a hash, as [preferred in its readme](https://github.com/ruby/setup-ruby#setup-ruby)

I'm happy to submit a PR for 1, 2, and/or 3. Just let me know which you'd like to see.

## Proof-of-concept for fix 1

As a proof-of-concept I created a minimal repository where CI fails, here:

https://github.com/jamiemccarthy/setup-ruby-test

The CI fails with:

```
Run bin/bundler-audit --update
  bin/bundler-audit --update
  shell: /usr/bin/bash -e {0}
/home/runner/work/_temp/ff391cbf-e7a3-41a6-8b3d-f46f381a4f9c.sh: line 1: bin/bundler-audit: No such file or directory
Error: Process completed with exit code 127.
```

and the `bundle exec` one-line fix makes the CI pass, here:

https://github.com/jamiemccarthy/setup-ruby-test/pull/1

```
Run bundle exec bundler-audit --update
  bundle exec bundler-audit --update
  shell: /usr/bin/bash -e {0}
Download ruby-advisory-db ...
Cloning into '/home/runner/.local/share/ruby-advisory-db'...
ruby-advisory-db:
[...]

PR #806, opened 2021-02-12
Clarify pre-requisites, tasks for onboarding of new starter workflows

swinton:
Proposes tune-ups to the existing PR template. The intent is to clarify requirements for proposing new starter workflows, with a shared section for common requirements, and dedicated sections for each "flavor" of workflow (CI, code scanning, ...).

For new Code Scanning starter workflows, we are requesting membership in our recently relaunched Partner Program.

PR #2302, opened 2024-02-12
Fix ci/rubyonrails bugs, using bundle exec

jamiemccarthy:
This PR fixes three separate issues I identified with the rubyonrails workflow in #2159:

1. Most seriously, the three linting commands must be run with bundle exec as a prefix, or they simply do not work. As written, the first lint command fails with `bin/bundler-audit: No such file or directory`. So this starter workflow has been broken for some time. This PR fixes that problem.
2. `bundle-audit` is the preferred spelling of that command, as [documented in its readme](https://github.com/rubysec/bundler-audit#readme)
3. ~~The [setup-ruby readme](https://github.com/ruby/setup-ruby#setup-ruby) notes: "Important: Prefer ruby/setup-ruby@v1" and this PR switches to pin to that tag instead of a commit. The readme explains "If you pin to a commit or release, only the Ruby versions available at the time of the commit will be available." The existing file, by pinning to a commit, has not only been missing [ruby 3.3.0](https://github.com/ruby/setup-ruby/commit/b256bd96bb4867e7d23e92e087d9bb697270b725), but has required manual updates to fix CVEs, as in eeb9248ea6d958f2adcb504c334b98464be041ea. It's more secure to let the upstream maintainers pick up these fixes and trust them to keep `v1` up to date.~~ _edit 2024-04-26: removed this change, see PR discussion for why_

I'm happy to discuss these changes, and to reformat them in whatever way will be easiest. In particular, I've pushed the changes above as separate commits but I'm happy to squash them and resubmit.

I've been using a modified form of this workflow in a couple of jobs/workflows in a personal project since September. For what it's worth, it uses the `bundle exec` change, and it's been working fine.

Thank you!


jsoref (commented):
cc: @indirect, @bishal-pdMSFT, @JamesMGreene

fwiw, these changes look reasonable to me.

indirect:
You can create the expected binstubs by running `bundle binstubs bundle-audit brakeman rubocop` and then checking in the generated files. I personally use binstubs rather than bundle exec, because I like the discoverability of seeing what's inside `bin/` in my application.

As for the `ruby-setup` version, I originally used v1 in my PR and was told that it is against policy for starter actions to be unpinned, and it was required to pin them to a precise hash for the PR to be accepted.

Good luck!

jsoref:
The pinning requirement is listed under **Previous guidelines for new starter workflows**:
https://github.com/actions/starter-workflows/blob/2435e576016b3fdd46574a49126d65bfd0d4872f/CONTRIBUTING.md?plain=1#L14-L22

It doesn't appear to be the actual policy for this repository

<details><summary>organizations/users whose repositories aren't pinned</summary>

* aliyun
* aws-actions
* azure
* docker
* github
* google-github-actions
* hashicorp
* kentaro-m
* microsoft
* peter-evans
* pozil
* redhat-actions
* slsa-framework
* synopsys-sig
* TencentCloud

</details>

I don't know what "Previous" means. It might mean "no longer in effect".

jamiemccarthy:
Ah, I see the requirement. Thanks for the heads-up, @indirect.

It's not in `CONTRIBUTING.md`, it's in the pull request template:

https://github.com/actions/starter-workflows/blob/2435e576016b3fdd46574a49126d65bfd0d4872f/.github/pull_request_template.md?plain=1#L53

> This workflow must _only_ use actions that are produced by the language or ecosystem that the workflow supports.  These actions must be [published to the GitHub Marketplace](https://github.com/marketplace?type=actions).  We require that these actions be referenced using the full 40 character hash of the action's commit instead of a tag.

This policy was adopted in 2021. At that time, there was discussion suggesting it shouldn't be a permanent restriction for some actions:

https://github.com/actions/starter-workflows/pull/806#pullrequestreview-589899124

> You can use a qualified partners workflow too. And they don't need pinning to a 40 char SHA

> Not sure how users will know which actions are from qualified partners... I'll merge this PR as-is for now, we can alway return to this and relax our conditions further in a future iteration.

I can see arguments either way, but that's out of scope for this PR.

So unless y'all object, I'll go ahead and push a commit that switches back from `@v1` to the latest hash.

I think the rest of this PR is still worth doing. I don't much care whether it uses `bundle binstubs` to write out new files or execs them with `bundle exec`. As long as the result is that it works :)

jamiemccarthy:
I don't have a strong preference for whether the ruby commands get run with `bundle exec` or as binstubs. This PR runs them with `bundle exec`. I've just pushed #2427 which runs them with binstubs.

I am proposing that one of them should be merged, depending on personal preference for how those ruby commands get run. The other PR should be closed. I don't care which 😄

jamiemccarthy:
@indirect @jsoref Just re-upping this request. This PR's been updated, and #2427 has also been submitted as an alternative that uses binstubs.

jsoref:
I have no influence over this repository. I have a project I want added to it, but I haven't even tried to submit it. I'm looking over existing PRs before I try to submit mine.

(I have made some PRs to bump generic dependencies, to test the waters, and it hasn't been particularly fast.)

indirect:
@jamiemccarthy I have gotten one PR merged here in the past. It took 3 months. Good luck!

jamiemccarthy:
Okay, thanks, sorry for the misunderstanding. I don't know if there's anyone to @-ping, so I'll just wait I guess!

thyeggman:
Closing this in favor of https://github.com/actions/starter-workflows/pull/2427

PR #2345, opened 2024-03-19
ci/python-publish: bump, use trusted publishing

### LINKED ISSUE #2344: Update python-publish.yml to use Trusted Publishing
I'm filing this for myself to work on 🙂 

Per https://github.com/github/docs/pull/32146 and the earlier https://github.com/github/docs/pull/24772: PyPI now supports authentication via OIDC ("Trusted Publishing") and other official GitHub resources mostly refer to that authentication mechanism.

This workflow should be similarly updated to use Trusted Publishing instead of a manually configured API token.

woodruffw:
Fixes #2344.

## Tasks

**Some general notes:**

    ```
    # documentation.
    ```


woodruffw:
xref https://github.com/actions/starter-workflows/issues/2344

CC @di @jhutchings1 

CC @webknjaz as well, as the maintainer of `gh-action-pypi-publish` 🙂

di (approved):

ci/python-publish.yml
```diff
@@ -12,28 +12,42 @@ on:
   release:
     types: [published]
 
-permissions:
```
webknjaz, line 15: It might be good to keep this globally in case people start modifying the jobs post workflow creation.

ci/python-publish.yml
```diff
@@ -12,28 +12,42 @@ on:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
 
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+
+      - name: build release distributions
```
webknjaz, line 26:
```suggestion
      - name: Build release distributions
```
(same)

ci/python-publish.yml
```diff
@@ -12,28 +12,42 @@ on:
...
+        run: |
+          # NOTE: put your own distribution build steps here.
+          python -m pip install build
+          python -m build
```
webknjaz, line 30: *(this is probably controversial but)* the CI images come with pipx pre-provisioned so this could just use that:
```suggestion
          pipx run --spec build python -Im build
```

ci/python-publish.yml
```diff
@@ -12,28 +12,42 @@ on:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+  pypi-publish:
     runs-on: ubuntu-latest
+    needs:
+      - release-build
+    permissions:
```
webknjaz, line 42: It'd *really* want to get people to use environment-based protection from the very beginning:
```suggestion
    environment:
      name: pypi
      url: https://pypi.org/p/<project-name>
    permissions:
```

ci/python-publish.yml
```diff
@@ -1,4 +1,4 @@
-# This workflow will upload a Python Package using Twine when a release is created
+# This workflow will upload a Python Package to PyPI when a release is created
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#publishing-to-package-registries
```
webknjaz, line 2: If it's not against some policy, let's link https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ — we're trying to keep it up-to-date over in PyPUG, which often happens sooner than GH gets docs updated.

ci/python-publish.yml
```diff
@@ -12,28 +12,42 @@ on:
...
+        with:
+          name: release-dists
+          path: dist/
+
+  pypi-publish:
     runs-on: ubuntu-latest
+    needs:
+      - release-build
+    permissions:
+      id-token: write
```
webknjaz, line 43: explanatory comment plz?

ci/python-publish.yml
```diff
@@ -12,28 +12,42 @@ on:
...
-        user: __token__
-        password: ${{ secrets.PYPI_API_TOKEN }}
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+      - name: Publish release distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
webknjaz, line 53: This might be a good place to give out a hardening recommendation:
```suggestion
        uses: pypa/gh-action-pypi-publish@release/v1  # update `release/v1` to a commit SHA if you use Dependabot
```

webknjaz:
@woodruffw love that this is moving somewhere! Thank you for getting to this sooner than me :) I've been frustrated with how many people get pre-historic workflows by default and don't even know it...

One extra thing to consider — it might be useful to also stick Sigstore signing right into the starter. OTOH, giving people a link to the guide might be an alternative.

ci/python-publish.yml
```diff
@@ -12,28 +12,42 @@ on:
...
-      with:
-        user: __token__
-        password: ${{ secrets.PYPI_API_TOKEN }}
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+      - name: Publish release distributions to PyPI
```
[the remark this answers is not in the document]
Plentywolf: Master script??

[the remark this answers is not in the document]
Plentywolf: 52 yes

webknjaz:
> I've been in discussion with @woodruffw, @di and the PyPI team about this.

Linking some context: https://github.com/pypa/gh-action-pypi-publish/pull/123#issuecomment-1982201999

ci/python-publish.yml
woodruffw: Makes sense, I'll re-add!

ci/python-publish.yml
woodruffw: Makes sense to me. I'll defer to GH folks for any actual policy here 🙂

ci/python-publish.yml
woodruffw: I'm -0 on this, for three reasons:

1. Fewer users know about `pipx`, so putting it here increases (very, very slightly) the cognitive complexity of the sample workflow. I recognize that this is a self-defeating argument, however, since this is the kind of thing needed to raise user awareness of `pipx` 🙂 
2. IMO `pipx run --spec` is a lot of indirection to introduce here: it requires users to think about `pipx`, about a venv for `build`, about a subprocess spawning `python` in that venv, etc. It's simpler for people like us, but I think this workflow also needs to cater to first-time publishers.
3. `pipx`'s availability on different runner images seems mixed/not subject to automated checks: https://github.com/actions/runner-images/issues/7590. Most users probably won't hit this since they'll be on `ubuntu-latest`, but it's another edge case we can dodge.

ci/python-publish.yml
woodruffw: Agreed, although this will produce an incongruity between this workflow and the other GitHub docs. I suppose we should update all of them to suggest using an environment.

ci/python-publish.yml
woodruffw: No objection, although I'm not sure if there's precedent in the starter workflows for this. I'll look.

ci/python-publish.yml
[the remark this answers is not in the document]
woodruffw: This might be a dialect thing: I've only ever heard "PyPI" without the *the* definite article, whereas expanded it would be "*the* Python Package Index" 🙂

ci/python-publish.yml
di: I agree.

woodruffw:
> One extra thing to consider — it might be useful to also stick Sigstore signing right into the starter. OTOH, giving people a link to the guide might be an alternative.

I'd personally like to shy away from suggesting the Sigstore action here for now, if only because (with PEP 740) it'll become obsolete and integrated directly into the publishing flow 🙂

ci/python-publish.yml
webknjaz: All the docs and examples I could reach already use environments called `pypi` and `testpypi` by default. Not `release` since that name is ambiguous and doesn't represent a specific deployment target (the PyPI).

webknjaz: GH docs you contributed earlier seem to use the `release` environment (so does the placeholder in Warehouse, which I dislike for the same reason).

ci/python-publish.yml
webknjaz: Okay, I withdraw this idea.

ci/python-publish.yml
webknjaz: My understanding is that they usually demand commit SHA in all starter workflows which is what got us into this mess in the first place. So I figured this might be a good middle ground 🤷‍♂️

woodruffw: Yeah, looks like `gem-push.yml` has a commit SHA, and a comment suggesting a change to a symbolic ref for automatic changes. I'll do something similar here 🙂

woodruffw: Did a version of what `gem-push.yml` does, PTAL!

ci/python-publish.yml
woodruffw: I was thinking of [this example](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-pypi#updating-your-github-actions-workflow), which (regrettably) doesn't use an environment. I'll add an environment here and work on updating that doc separately.

woodruffw: Added an environment example with some comments, PTAL!

ci/python-publish.yml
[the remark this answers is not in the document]
webknjaz: It's a grammar thing (think: *the* US). Though I indeed saw various usage patterns as well. OTOH PyPI is unique enough not to need an article (people don't tend to confuse it with Test PyPI).

I'm likely confusing it with the preferred usage of "the PyPA". I did some googling and only found it the long form: https://packaging.python.org/en/latest/glossary/#term-pypi.org. The FAQ on pypi.org doesn't use the article in front of the abbreviation.

https://english.stackexchange.com/a/82827/68752 suggests this depends on how one would pronounce an abbreviation.

Anyway, I'm going to trust your instinct. Please, disregard this suggestion.

ci/python-publish.yml
webknjaz: Personally, I prefer the rolling branch but I don't think the GH folks would agree to that.. Dependabot wouldn't update it, though (until converted to a tag or a commit hash).
I suppose your variant will work for most.

ci/python-publish.yml
webknjaz: Looks good. Normally I like making versioned URLs. But that might be confusing to newbies. Although, if there's an easily retrieveable version in the release event, sticking it into the example comment would be a nice touch.

woodruffw: I think there is, although there's no guarantee that the release version is a match for the PyPI version (user error, but also because of PEP 440 normalization/rewrites). So I'm inclined to leave this unversioned for now, although I don't have a super strong opinion here 🙂

webknjaz: I think it's fine if it's not normalized and the URL might be off (although, personally I started embedding `workflow_dispatch` inputs validation for PEP 440), and if it's in a comment, as a hint. The most reliable way of getting that would be unpacking a wheel or sdist and grabbing it from the structured metadata since the packaging machinery usually normalizes it already. I won't insist, though.

woodruffw:
This should be good to go. I think it might make sense to review and land this before https://github.com/github/docs/pull/32146, since the workflow changes here will need to be reflected there as well.

woodruffw:
Gentle ping for review here!

(@jhutchings1 I'm calling in that promise 😉)

woodruffw:
Another gentle ping for review here: I believe this should be merged before https://github.com/github/docs/pull/32146 goes forwards, since those docs will link to this workflow.

jhutchings1:
@nebuk89 Can you help get this prioritized?

ci/python-publish.yml
```diff
@@ -16,24 +16,51 @@ permissions:
...
+        # To automatically get updates, change this to:
+        # uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@81e9d935c883d0b210363ab89cf05f3894778450 # v1.8.14
```
pandafy, line 66: There is a newer release of the action available. 

```suggestion
        uses: pypa/gh-action-pypi-publish@ec4db0b4ddc65acdf4bff5fa45ac92d78b56bdf0 # v1.9.0
        
```

woodruffw:
Gentle ping for review here -- it'd be _really_ nice to get this landed, since other parts of GitHub's docs now refer extensively to PyPI's support for OIDC/Trusted Publishing.

pandafy:
It took me sometime to debug why trusted publishing on my package was not working, and it was due to the old commit hash present in the starter code. Getting this merged would be wonderful.

ci/python-publish.yml
```diff
@@ -16,24 +16,52 @@ permissions:
...
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+
+    # Dedicated environments with protections for publishing are strongly recommended.
+    # For more information, see: https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#deployment-protection-rules
+    environment:
+      name: pypi
+      # OPTIONAL: uncomment and update to include your PyPI project URL in the deployment status:
+      # url: https://pypi.org/p/YOURPROJECT
```
webknjaz, line 54: @woodruffw how about mentioning that versioned URL as optional?
```suggestion
      # url: https://pypi.org/p/YOURPROJECT
      #
      # ALTERNATIVE: if your GitHub Release name is the PyPI project version string
      # ALTERNATIVE: exactly, uncomment the following line instead:
      # url: https://pypi.org/project/YOURPROJECT/${{ github.event.release.name }}
```

jhutchings1:
@actions/actions-workflow-development-reviewers Can we get a PR review on this? This is the best practice from the ecosystem maintainers and this PR has been sitting for a very long time. cc: @nebuk89 @chrispat @steiza

webknjaz:
@woodruffw perhaps, we could stick attestations into this, finally?

woodruffw:
> @woodruffw perhaps, we could stick attestations into this, finally?

Do we need to do anything explicitly for them here? The way we've designed the attestation support in `gh-action-pypi-publish` means that they'll become the default once we switch the flag, so (IIUC) we should be able to leave this workflow mostly as-is and users will begin attesting as they upgrade to the version that makes it the default.

That being said, it probably makes sense to update the version/commit ref in this PR once again. I'll do that now.

ci/python-publish.yml
```diff
@@ -16,24 +16,55 @@ permissions:
...
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+      - name: Publish release distributions to PyPI
+        # To automatically get updates, change this to:
+        # uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@0ab0b79471669eb3a4d647e625009c62f9f3b241 # v1.10.1
```
webknjaz, line 70:
```suggestion
        uses: pypa/gh-action-pypi-publish@fb13cb306901256ace3dab689990e13a5550ffaa  # v1.11.0
```

woodruffw:
Gentle ping -- this has been sitting for a while now and is now behind the other GitHub docs (and PyPI/PyPUG docs) in terms of what it contains.

CC @jhutchings1 @nebuk89

jhutchings1:
I'm really bummed this one hasn't gotten traction. I'm not an employee anymore, so don't have any official seat, but @chrispat  anything you can do to get this one some love?

jhutchings1:
I'm really bummed this one hasn't gotten traction. I'm not an employee anymore, so don't have any official seat, but @chrispat  anything you can do to get this one some love?

ci/python-publish.yml
```diff
@@ -16,24 +16,55 @@ permissions:
...
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+      - name: Publish release distributions to PyPI
+        # To automatically get updates, change this to:
+        # uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@fb13cb306901256ace3dab689990e13a5550ffaa  # v1.11.0
```
webknjaz, line 70:
```suggestion
        uses: pypa/gh-action-pypi-publish@1f5d4ec244f65dce93685ee3e98e77123f090866  # v1.12.1
```

ci/python-publish.yml
```diff
@@ -16,24 +16,55 @@ permissions:
...
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+      - name: Publish release distributions to PyPI
+        # To automatically get updates, change this to:
+        # uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@1f5d4ec244f65dce93685ee3e98e77123f090866  # v1.12.1
```
webknjaz, line 70:
```suggestion
        uses: pypa/gh-action-pypi-publish@15c56dba361d8335944d31a2ecd17d700fc7bcbc  # v1.12.2
```

ci/python-publish.yml
```diff
@@ -16,24 +16,55 @@ permissions:
...
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+      - name: Publish release distributions to PyPI
+        # To automatically get updates, change this to:
+        # uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@15c56dba361d8335944d31a2ecd17d700fc7bcbc  # v1.12.2
```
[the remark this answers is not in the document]
steiza: Separately: in my test case I had a custom directory. This worked fine, after I found `packages-dir`. Even though it's the default, should we provide `dist/` here like we do in the above commands?

I could go either way.

```suggestion
        uses: pypa/gh-action-pypi-publish@15c56dba361d8335944d31a2ecd17d700fc7bcbc  # v1.12.2
        with:
          packages-dir: dist/
```

steiza (approved):
3 minor comments, otherwise looks good to me! I was able to use this workflow successfully.

ci/python-publish.yml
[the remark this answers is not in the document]
woodruffw: +1 to making it explicit and pinning on `release/v1` instead of a digest -- normally I'd say we should suggest the digest, but the long review turnaround here indicates that doing so will be a maintenance/staleness risk 🙂

ci/python-publish.yml
```diff
@@ -16,24 +16,57 @@ permissions:
...
```
woodruffw, line 70:
```suggestion
        uses: pypa/gh-action-pypi-publish@release/v1
```

woodruffw:
Thanks @steiza, applied!

steiza (approved):
Fantastic; thanks! I think this is ready to merge.

woodruffw:
Thanks a ton @steiza and @elbrenn!

webknjaz:
Thanks @elbrenn and @steiza!

PR #2390, opened 2024-04-30
go: update to use go1.22

chenrui333:
go 1.20 is EOL, update to use 1.22

also update setup-go to use nodejs 20 runtime dep


mScott224 (approved):

mScott224 (approved):

JamesonDiamond1 (approved):
@JamesonDiamond1

PR #2412, opened 2024-05-29
CodeQL: Add `actions: write` permission to support TRAP cache cleanup

henrymercer:
This PR adds the `actions: write` permission to the CodeQL workflow so that the CodeQL workflow can automatically cleanup unused CodeQL TRAP caches.

While we're here, we also add a step name for the manual build commands step, so it's more clear what is going on when this step is skipped.

### DIFF
--- a/code-scanning/codeql.yml
+++ b/code-scanning/codeql.yml
@@ -77,7 +77,8 @@ jobs:
     # to build your code.
     # ℹ️ Command-line programs to run using the OS shell.
     # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
-    - if: matrix.build-mode == 'manual'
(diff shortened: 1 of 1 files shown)


aeisenberg:
As we discussed, is there a way to restrict `actions: write` to only be added for push events?

marcogario (approved):

jsoref:
@marcogario / @thyeggman is someone going to merge this?

henrymercer:
Hey @jsoref — this work is currently on pause.  We are leaning towards wanting to limit the permissions of jobs that might run build commands to `actions: read` permissions, so when we resume it we'll take a different approach to TRAP cache cleanup.  Specifically, we are thinking of running cleanup in a separate job.  We can then give only that cleanup job `actions: write` permissions.  I'll close the PR accordingly.

PR #2466, opened 2024-08-06
Update soos-dast-scan.yml hash

SOOS-GSteen:
Update soos-dast-scan.yml hash


alexisabril (approved):
Thank you @SOOS-GSteen!

PR #2597, opened 2024-11-13
Remove excess permissions from pages workflows

jsoref:
The various pages workflows all set a top level permissions block which was unfortunate.

With the exception of the static workflow, each workflow had two jobs, one that only needed `contents: read` and one that didn't even need `contents: read`.

I'm splitting the permissions block for each of those workflows such that the build phase only gets `contents: read` and the publish phase gets `pages: write` and `id-token: write` (but not even `contents: read` as it doesn't need it).

For the static workflow, I'm moving the permissions into the job for consistency and also to make it easier to compose this job into some bigger workflow.

I discovered the overly generous permissions when I was applying the pages/jekyll-gh-pages.yml into a workflow I have...

---

## Tasks

- [ ] Should be contained in a `.yml` file with the language or platform as its filename, in lower, [_kebab-cased_](https://en.wikipedia.org/wiki/Kebab_case) format (for example, [`docker-image.yml`](https://github.com/actions/starter-workflows/blob/main/ci/docker-image.yml)).  Special characters should be removed or replaced with words as appropriate (for example, "dotnet" instead of ".NET").
- [ ] Should use sentence case for the names of workflows and steps (for example, "Run tests").
- [ ] Should be named _only_ by the name of the language or platform (for example, "Go", not "Go CI" or "Go Build").
- [ ] Should include comments in the workflow for any parts that are not obvious or could use clarification.
- [ ] Should specify least privileged [permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) for `GITHUB_TOKEN` so that the workflow runs successfully.

**Some general notes:**

    ```
    # documentation.
    ```

### DIFF
--- a/pages/astro.yml
+++ b/pages/astro.yml
@@ -12,12 +12,6 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
-# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
-permissions:
-  contents: read
-  pages: write
-  id-token: write
-
 # Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
 # However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
 concurrency:
@@ -31,6 +25,9 @@ env:
 jobs:
   build:
     name: Build
+    permissions:
+      contents: read
+
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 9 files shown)


mScott224 (approved):

JamesonDiamond1 (approved):
@JamesonDiamond1

jsoref:
@JamesMGreene can someone please look at this?

JamesMGreene:
@jsoref I don't have the permissions to approve/merge anymore, but I'll reach out to my old team to see if anyone can get to it soon.

PR #2639, opened 2024-11-29
Bump python app

sicheste:
Using setup-python v5 and Python 3.13.

## Pre-requisites

---

---

## Tasks

- [ ] Should be preserved under [the `code-scanning` directory](https://github.com/actions/starter-workflows/tree/main/code-scanning).
- [ ] Should include a matching `code-scanning/properties/*.properties.json` file (for example, [`code-scanning/properties/codeql.properties.json`](https://github.com/actions/starter-workflows/blob/main/code-scanning/properties/codeql.properties.json)), with properties set as follows:
  - [ ] `name`: Name of the Code Scanning integration.
  - [ ] `creator`: Name of the organization/user producing the Code Scanning integration.
  - [ ] `description`: Short description of the Code Scanning integration.
  - [ ] `categories`: Array of languages supported by the Code Scanning integration.
  - [ ] `iconName`: Name of the SVG logo representing the Code Scanning integration. This SVG logo must be present in [the `icons` directory](https://github.com/actions/starter-workflows/tree/main/icons).
- [ ] Should run on `push` to `branches: [ $default-branch, $protected-branches ]` and `pull_request` to `branches: [ $default-branch ]`. We also recommend a `schedule` trigger of `cron: $cron-weekly` (for example, [`codeql.yml`](https://github.com/actions/starter-workflows/blob/c59b62dee0eae1f9f368b7011cf05c2fc42cf084/code-scanning/codeql.yml#L14-L21)).

**Some general notes:**

    ```
    # documentation.
    ```

### DIFF
--- a/ci/python-app.yml
+++ b/ci/python-app.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python 3.10
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v5
       with:
         python-version: "3.10"
     - name: Install dependencies
--- a/ci/python-app.yml
+++ b/ci/python-app.yml
@@ -19,10 +19,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python 3.10
+    - name: Set up Python 3.13
       uses: actions/setup-python@v5
       with:
-        python-version: "3.10"
+        python-version: "3.13"
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/ci/python-app.yml
+++ b/ci/python-app.yml
@@ -19,10 +19,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python 3.13
+    - name: Set up Python 3.12
       uses: actions/setup-python@v5
       with:
-        python-version: "3.13"
+        python-version: "3.12"
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip


ci/python-app.yml
```diff
@@ -19,10 +19,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python 3.10
-      uses: actions/setup-python@v3
+    - name: Set up Python 3.13
+      uses: actions/setup-python@v5
       with:
-        python-version: "3.10"
+        python-version: "3.13"
```
reactive-firewall, line 25: LGTM! with Conditions:

 * verified improvement of `actions/setup-python@v5` (from `v3`).
 * Meta-data is consistent and matches the Python version used.
   <details><summary>NIT</summary>

     > [!TIP]
     > knit-picky or out-of-scope:
     > Consider improving with a single variable to set the python version and
     > access it for both name and python-version via GHA contexts.
     > See [Documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs) for details.)

   </details>
   <details><summary>NIT - Consider Python `3.12` might be more appropriate than `3.13` for GHA</summary>

     > [!TIP]
     > knit-picky or out-of-scope:
     > Looks like version 3.12 will be the cached version (while `3.13` **is** the
     > latest it will probably be slower for some GitHub Actions)
     > see [https://github.com/actions/runner-images/issues/10636](https://github.com/actions/runner-images/issues/10636)
     > and note the cached versions for python.
     >
     > Should still work though; Just slightly slower.

   </details>

**Hope this helps** :bow:

reactive-firewall (approved):

mScott224 (approved):

ci/python-app.yml
cclauss, line 25:
```suggestion
        python-version: "3.x"
```
https://devguide.python.org/versions/#supported-versions
Testing on the latest version of Python may catch new language incompatibilities before most users notice them.

cclauss (approved):

ci/python-app.yml
reactive-firewall: > [!WARNING]
> 🙅🏻 Syntax error

@cclauss [that suggestion](https://github.com/actions/starter-workflows/pull/2639#discussion_r1874591190) would regress the action to a broken state. Please re-read the docs you correctly mentioned.

I'll also re-iterate what I previously marked knitpicky as it now directly applies to this, see below:

Consider Python `3.12` might be more appropriate than `3.13` for GHA
> [!TIP]
> knit-picky or out-of-scope:
Looks like version 3.12 will be the cached version (while 3.13 is the
latest it will probably be slower for some GitHub Actions)
see https://github.com/actions/runner-images/issues/10636
and note the cached versions for python.

> Should still work though; Just slightly slower.

In conclusion; pick one (exactly one):
 * `3.12` - for tuned value based on GitHub Action's runners.
 * `3.13` - the latest available version of Python as per your link; "Should still work though; Just slightly slower" due to not being cached.
 * any other less current value in the form defined by the [`actions/setup-pythonv@5`](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md) (`3.9.20`|`3.10`|`3.11`) (included for completeness)

> [!IMPORTANT]
> This is not to say your rational is wrong, (only the implementation)

To also implement checking for the latest version please refer to [this guide](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#check-latest-version)

or as code:
```python
# Rest of config is unchanged
  - uses: actions/setup-python@v5
    with:
      python-version: '3.13'
      check-latest: true
```

sicheste: I like the idea using a cached Python version by default. So I switched to 3.12.

sicheste: Regarding the variables:
I've checked some other files. Have not seen that it was done like this somehwere. For example the android, ant and maven templates are using the version number in the step name and for the version requirement. But always explicitly written. So yes, technically a considerable idea, but have not seen that it is done somewhere else like this.

Using Python 3.12 instead of 3.13 makes sense and I have changed.

sicheste: In python-publish it is given with `3.x`. So would not be the first template when setting it to .x here.

reactive-firewall: > In python-publish it is given with `3.x`. So would not be the first template when setting it to .x here.


## Oops 🙊

My oversight; Correction: it is a runtime error, not a syntax error. I'll remove the incorrect and confusing phrasing from my previous comment.

For those curious, my full rational,
<details><summary>TL;DR</summary>

This gave me pause, at first, as you are quite correct, many (e.g. [`python-publish`](https://github.com/actions/starter-workflows/blob/main/ci/python-publish.yml#L27), and even the GitHub guides use `3.x` all over, and only [sometimes do they document why](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#using-a-specific-python-version)). After some digging all I could find was:
```yaml
name: Python package

on: [push]

jobs:
  build:

    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4
      - name: Set up Python
        # This is the version of the action for setting up Python, not the Python version.
        uses: actions/setup-python@v5
        with:
          # Semantic version range syntax or exact version of a Python version
          python-version: '3.x'
          # Optional - x64 or x86 architecture, defaults to x64
          architecture: 'x64'
      # You can test your matrix by printing the current Python version
      - name: Display Python version
        run: python -c "import sys; print(sys.version)"
```

which makes sense if you are familiar with [Semantic Version range syntax](https://semver.org/#semantic-versioning-specification-semver) which boils down to [almost the same thing](https://regex101.com/r/vkijKf/1/) as I expected (e.g. PEP-440 and the current [Python packaging version spec](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers))
The notable difference is `3.x` is invalid as a **Semantic Version** but is considered a local version `x` of version with Major `3` resulting in 

so I checked [the code](https://github.com/actions/setup-python/blob/main/action.yml#L7C5-L7C150) which documents the same thing, (albeit without the example placeholder of `3.x`). So I wanted to see what it did with the version, and deep inside the massive [`dist/index.js`](https://raw.githubusercontent.com/actions/setup-python/refs/heads/main/dist/setup/index.js) of the `python-setup` action is a function `resolveVersionInput()` to resolve the input range that would essentially return `['3.x']`, from there it is used by the `run()` function to attempt to download it via the function `useCpythonVersion` (because `3.x` will not match other cases like 'PyPy' etc.) there it finally is parsed (and unless `python3.x` cpython is already installed, will fail and fall-back on a default of `void 0` (e.g. `None` version), if there is a `python3.x` installed on the GHA image is most likely the documented default of `3.12`) resulting eventually in a harmless runtime-error, and reporting the actually installed version via the output `python-version` (same name as input) which would currently be `3.12`.

In conclusion: `3.x` is still not advised, as omitting the value would be preferred to achieve the same behavior;
the original `3.13` should work;
and I still recommend the cached `3.12`.

### Informative references
 * `actions/setup-python@v5`/[`dist/index.js`](https://raw.githubusercontent.com/actions/setup-python/refs/heads/main/dist/setup/index.js) - the code.
 * [using-a-specific-python-version](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#using-a-specific-python-version) - GitHub guide on python-setup action topic
 * [python-action](https://github.com/actions/setup-python) and [docs](https://github.com/actions/setup-python#basic-usage)
 * [semantic versioning specification](https://semver.org/#semantic-versioning-specification-semver)
 * [semantic versioning reg-Ex](https://regex101.com/r/vkijKf/1/) - spec recommended way to test version format
 * [Python versioning specification](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers)
</details>

---

## Summary

In summary: This topic has been thoroughly explored now, I agree with @cclauss that just using the latest stable would be a fine alternative, but not better than the other already mentioned options.

@sicheste `3.12` for leveraging caching in GHAs, looks good, 👍🏻

ci/python-app.yml
reactive-firewall, line 1: **LGTM** 👍🏻

* consistent meta-data and values
* versions are stable and current

reactive-firewall (approved):

ci/python-app.yml
sicheste: Thanks for the long analysis and explanation.

So, now someone can merge it :)

ci/python-app.yml
JamesonDiamond1: @JamesonDiamond1

JamesonDiamond1 (approved):
@JamesonDiamond1

reactive-firewall (approved):
LGTM :+1:
 * versions are stable and current

JamesonDiamond1 (approved):
its not uplaoding a complete file

sicheste:
outdated

PR #2790, opened 2025-03-03
Add --locked to Rust cargo actions

tmandry:
This causes dependencies to be resolved in exactly the same way as they were in the Cargo.lock file, leading to more deterministic builds. Committing Cargo.lock and using this flag in CI is considered best practice, as suggested by the cargo book.

https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options

### DIFF
--- a/ci/rust.yml
+++ b/ci/rust.yml
@@ -17,6 +17,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Build
-      run: cargo build --verbose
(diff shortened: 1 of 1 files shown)


straight-flush (approved):
dreamverse

straight-flush:
dream verse infornation

JamesonDiamond1 (approved):
@JamesonDiamond1

PR #3205, opened 2026-02-13
Add SvelteKit workflow and properties for GitHub Pages deployment

Donny-C-1:
### Description
This PR adds a dedicated starter workflow for SvelteKit projects deploying to GitHub Pages. Currently, SvelteKit users often have to manually adapt generic Node.js or Next.js workflows, which leads to common pitfalls regarding `adapter-static` and build output directories.

### Key Features
- **Auto-Detection:** Detects `npm` vs `yarn` based on lockfiles.
- **Fail-Fast Validation:** Includes a pre-check to ensure `@sveltejs/adapter-static` is installed, preventing confusing build failures.
- **Optimized Caching:** Uses `actions/setup-node` caching for faster dependency resolution.
- **Dynamic Base Path:** Passes the repository name as `BASE_PATH` to handle SvelteKit's routing in project-site subfolders.

### Testing
- [x] Verified build output on a standard SvelteKit skeleton project.
- [x] Tested `npm ci` and `yarn install` pathways.
- [x] Confirmed successful deployment using the `actions/deploy-pages` action.

### Tasks


Montiwa11 (approved):

Montiwa11 (approved):
@Montiwa11

Mazafaker13:
[Uploading Add SvelteKit workflow and properties for GitHub Pages deployment by Donny-C-1 · Pull Request #3205 · actions_starter-workflows…]()

ev-max2024:
validate-dataExpected — Waiting for status to be reported

Donny-C-1:
What does this validate-data expected  - waiting for status to be reported mean ??
And why is it still blocking the pr

PR #3296, opened 2026-05-08
Updates 'setup-ruby' to v1 in the Jekyll Workflow Template

eliduke:
Recently started a new GitHub Pages / Jekyll project, and was surprised to see the `setup-ruby` action was still tied to a very old and specific version. Not sure why it's still held there, but I was getting some errors during my first couple deploys and I couldn't figure out what was going on!

I upgraded a bunch of my own personal projects to v1 a while back, and I had just assumed that the Jekyll workflow template would have been updated as well.

Thoughts? Concerns?


xoxoxo-af (approved):

thatrobotdev:
I had the same issues! The workflow is currently pinned to [`setup-ruby`v1207.0](https://github.com/ruby/setup-ruby/releases/tag/v1.207.0) which produces this error:

```
Using Bundler 4.0.1 from Gemfile.lock BUNDLED WITH 4.0.1
/opt/hostedtoolcache/Ruby/3.1.6/x64/bin/gem install bundler -v 4.0.1
ERROR:  Error installing bundler:
 bundler-4.0.1 requires Ruby version >= 3.2.0. The current ruby version is 3.1.6.
```

I tried upgrading incrementally:
* [`setup-ruby`v1.132.0](github.com/ruby/setup-ruby/releases/tag/v1.132.0) (added Ruby 3.2.0 support) still fails with `Error: Unsupported platform ubuntu-24.04` on `latest-ubuntu`.
* [`setup-ruby`v1.177.0](github.com/ruby/setup-ruby/releases/tag/v1.177.0) (added `ubuntu-24.04` support) works!

So the minimum fix seems to be pinning to v1.177.0. We should also explicitly pin the runner to `ubuntu-24.04` to prevent breakage in the future if we go this route.

eliduke:
Is there a need to pin it to a specific version beyond just `v1`? Doesn't that automatically pull the latest minor version as well?

evanwill:
I think that error isn't so much with the version of setup-ruby as it is with the version of ruby--note on next line `ruby-version: '3.1'`. ruby 3.1 is end of life, if you bump that to `3.4` the action still works fine. 

The action claims to support `.ruby-version` file but I haven't had success with it.

eliduke:
@evanwill Are you thinking to make some adjustments in this PR to address your issue #3329 ?

The line in question here...

https://github.com/actions/starter-workflows/blob/1035244887e26fbbd4f1017d919fb5995cc521c4/pages/jekyll.yml#L40

The way that I read that and the comment included is that you need to remove the *entire line* for it to read from your `.ruby-version` file. It works great for me like this...

```yaml
- name: Setup Ruby
  uses: ruby/setup-ruby@v1
  with:
    bundler-cache: true # runs 'bundle install' and caches installed gems automatically
    cache-version: 0 # Increment this number if you need to re-download cached gems
```

Thoughts? I think you are correct that we shouldn't have such an old version of ruby as the default here, but I'm not sure exactly what the default should be.

evanwill:
I have been using `ruby-version: '3.4'` as the fix, since 3.4 is the most recent in the 3 line. Jumping to 4 is a bit more likely to break things for people at this point I think. 

Thanks on the .ruby-version file clarification--I just always figured if I had it .ruby-version it should override what is in the action, not that I would also have to comment it out!

PR #3361, opened 2026-07-08
Add Scrutora Scan code-scanning starter workflow

Nirvahana:
Adds a starter workflow for **Scrutora Scan** under `code-scanning/`.

Scrutora Scan is an offline, compliance-mapped SAST tool (DPDPA, HIPAA, GDPR, PCI-DSS, RBI). It requires **no API key** and code never leaves the runner; results upload to GitHub Code Scanning as SARIF via `github/codeql-action/upload-sarif`.

**Files**
- `code-scanning/scrutora-scan.yml` — the workflow (uses `Scrutora/scrutora-scan@v1`)
- `code-scanning/properties/scrutora-scan.properties.json` — metadata
- `icons/scrutora.svg` — icon

Follows the third-party contribution format (based on the existing `bearer` workflow): includes the "not certified by GitHub" header, `$default-branch`/`$cron-weekly` placeholders, least-privilege `permissions`, and `fail-on: none` so it reports without gating the build by default.

Action: https://github.com/Scrutora/scrutora-scan


Nirvahana:
Thanks for the review — all four addressed:

1. **Branch placeholders** — switched to unquoted `$default-branch` and added `$protected-branches` on `push`, matching the repo convention (e.g. `bearer.yml`).
2. **RBI claim (workflow header)** — removed; the header now lists only DPDPA/HIPAA/GDPR/PCI-DSS, consistent with the action's documented frameworks.
3. **RBI in metadata description** — removed for the same reason.
4. **Hard-coded SARIF filename** — the scan step now has `id: scan` and the upload uses the action's output: `sarif_file: ${{ steps.scan.outputs.sarif-file }}`.

Both files re-validated. Let me know if anything else is needed.

